PR #36, opened 2014-09-10
Implements the new synchro of team members from odoo to GitHub

guewen:
As discussed with @gurneyalex, the project teams will no longer hold all the
core committers. There will be an additional team for the core committers
assigned to each project. The edition of the team members will be done on odoo,
never on GitHub.

What will do this script:

* It considers odoo as the master, meaning that, for each team,
  it removes from GitHub all the members that are not members in
  odoo. Conversely, it adds the new members who have been added on odoo.
* It reports the users who should be added to a team but can't because their
  GitHub login is missing.
* Teams are matched using their name, it reports the teams not found on GitHub.

One side effect of the first statement is that odoo users who have not their
GitHub login set on odoo will be removed from the GitHub teams, so the first thing
to do is to fill in all the logins. Users that miss their login actually:

* [x] Sylvain LE GAL
* [x]  Joël Grand Guillaume
* [x]  Eric Caudal
* [x]  Sandy Carter
* [x]  Rijnhart Stefan
* [x] ~~admin~~
* [x]  Nicolas Bessi
* [x]  Nhomar Hernández
* [x]  Sébastien Beau
* [x]  Vaucher Yannick
* [x]  Maxime Chambreuil

You will note the presence of the admin user, we could blacklist it but I
prefer to leave it because it should be removed from all the teams anyway.

Then, we must ensure that team names match on both sides. The odoo teams not
found on GitHub:

* [x]  Connector CMIS Maintainers
* [x]  Local Belgium Maintainers
* [x]  Local Luxemburg Maintainers
* [x]  Local Mexico Maintainers
* [x]  Product Maintainers
* [x]  Project & Services Maintainers

Other things to do:
* [ ] We'll also need to have a new team "Core Maintainers" or "Core
Committers" assigned on all the OCA projects.
* [ ] and add / check the members in the odoo teams
* [ ] We probably want to setup a cron somewhere.


gurneyalex:
Summing up discussion today with @guewen and @StefanRijnhart:

* @gurneyalex to create a script to sync create res.users on the OCA instance for all members with a github login filled in -> this will lift the blocking part of this PR
* @StefanRijnhart to review the PR and propose enhancements if required. 

Then we merge and move on. 

In the future when https://gist.github.com/chrissiebrodigan/4c268535f0341c96016d is available in github, we may change the flow to manage things in github and synchronize back to the OCA instance.

StefanRijnhart:
Trying to test this with --dry-run but the erppeek bit fails to login to the OCA instance with my email address and password. Any known gotchas with this?

guewen:
> Trying to test this with --dry-run but the erppeek bit fails to login to the OCA instance with my email address and password. Any known gotchas with this?

Oh yes, a password must be set on your user on the instance: https://www.odoo.com/forum/help-1/question/is-the-xml-rpc-api-available-on-openerp-online-915#answer_937

StefanRijnhart:
Thanks! @gurneyalex has now kindly provided me with an API password for my account.

StefanRijnhart:
Works like a charm. I dry runned all the repositories and synched the Core Maintainers team. I also corrected some inconsistencies between the OCA instance and GitHub. :+1: for the merge but we would need to add users to all github-enabled partners in the OCA instance first, using #47 and give the project's OCA representatives (aka. project leaders) the opportunity to amend their project members in the instance before the script throws everyone out of the GitHub teams. Thank you so much for your work!

jgrandguillaume:
+1 from OCA board point of view

guewen:
merged with #47

PR #42, opened 2014-11-14
[ADD] create pr script

moylop260:


gurneyalex:
@moylop260 why cook our own when we can use https://github.com/github/hub?

dreispt:
@moylop A challenge :sunrise: : what about writing a Python Hub-like tool? I've been looking for that but it's yet to be created.
That script could server as a first version. It would go on it's won repo, outside OCA.

gurneyalex:
@dreispt Have you looked at the Python libraries listed on https://developer.github.com/libraries/?

dreispt:
@gurneyalex No I haven't, thanks for the tip.

moylop260:
Hello,
I tried unsuccessfully to use libraries

Then I copied this section code from runbot:
https://github.com/odoo/odoo-extra/blob/master/runbot/runbot.py#L224-L243

And later I use this method:
```python
    pr_data = {
        "title": title,
        "head": branch_head,
        "base": branch_base,
        "body": comment,
    }
    return github(repo_base=project_base,
                  url='/repos/:owner/:repo/pulls', payload=pr_data)
```
I don't have problem if we use library, but I don't found how use it to create a pull request.

guewen:
IMO, the `maintainers-tools` project has not for purpose to reimplement `hub` in Python.
Why not improve an existing Python project or use `hub`?

lmignon:
I agree with @guewen, IMO this functionality  should be provided by an existing python lib. If it doesn't exist, it would be nice to improve an existing lib. One common mistake in OSS is to reinvent in place of contribute..

yvaucher:
I'll close this PR as the main opinion is that's out of scope of maintainer tools repository

PR #44, opened 2014-12-28
[ADD] Description templates (rst, html)

max3903:
Provide templates in rst and html format for module description to be displayed on Odoo Apps, an Odoo instance or Github (rst only).


gurneyalex:
:+1:

PR #46, opened 2015-01-27
Add standards and guidelines

dreispt:
OCA needs documentation on coding standards and on the contributing and reviewing processes.

There is some documentation on the [OCA website](http://odoo-community.org/page/Contribute), but I believe that GitHub is a better place to discuss and evolve those rules.

My proposal is to (re)use this project to host that documentation.
We could have the other project's README files point here with a "How to contribute" link.


yvaucher:
Shouldn't we put that in wiki with a link to the wiki instead?

https://github.com/OCA/maintainers-tools/wiki

dreispt:
With In repo files we can use PRs to discuss and control changes to the guidelines.
IMO that's an important capability.

dreispt:
I would like to open the discussion on this.
IMO the [current guidelines](https://doc.odoo.com/contribute/05_developing_modules/#community-review) should be moved here, into docs in the repo, not wiki.
This allows opening PRs to propose and discuss changes to the guidelines.

Thoughts?

bwrsandman:
I agree with @yvaucher that a wiki should be used, but I honestly don't think maintainer should be the repo for it.

My reasoning is that this is a maintainer repository and the average user is not going to be looking at this repo.

It's my opinion that the guidelines should be somewhere more accessible.

A good compromise would be to have a guarantee that, from the start, wherever on github we put the guidelines, the OCA site updates its content automatically to reflect what has been agreed upon on github.

bealdav:
+1 for PR on this matter

guewen:
The weird thing is that this repo is an egg initially and this readme is also the description of the egg: https://github.com/OCA/maintainer-tools/blob/master/setup.cfg#L6

yvaucher:
@bwrsandman Didn't you worked on this topic on last code sprint? Can we close this?

bwrsandman:
I did. It can be closed

yvaucher:
Then it's closed

@dreispt You can see the result here https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md a look at it would be welcome.

PR #47, opened 2015-02-03
Create users in OCA instance for partners with a github login

gurneyalex:


gurneyalex:
This PR depends on #36  (and I actually rebased my branch on top of that one)

@StefanRijnhart @guewen your feedback is welcome

gurneyalex:
@StefanRijnhart fixed thanks.

tools/oca_sync_users.py
```diff
@@ -0,0 +1,38 @@
...
+def main():
+    parser = get_parser(with_help=True)
+    args = parser.parse_args()
+    client = login(args.username, args.store)
+    ResPartner = client.ResPartner
+    ResUsers = client.ResUsers
+    ResGroups = client.ResGroups
+    grp_project_user = ResGroups.get('project.group_project_user')
+    members_with_gh = ResPartner.search([('x_github_login', '!=', False),
+                                         ('user_ids', '=', False)])
```
StefanRijnhart: Maybe check for active_test here, so that we don't try to recreate users which have been set to inactive for some reason?

tools/oca_sync_users.py
```diff
@@ -0,0 +1,38 @@
...
+    for partner in ResPartner.browse(members_with_gh):
+        try:
+            user = ResUsers.create({'partner_id': partner.id,
+                                    'login': partner.email,
+                                    'groups_id': [(4, grp_project_user.id, 0)],
+                                    })
+        except xmlrpclib.Fault as exc:
+            print('unable to create user for partner %r: '
+                  'missing email address' % partner.x_github_login)
```
StefanRijnhart: You can probably have an exception here if there already is a user with the same email address so maybe you want to rephrase, or attempt to match existing users with this email address in an earlier stage.

gurneyalex:
@StefanRijnhart the current problem I have is with duplicate partners: one has a github login set, no email, no user, the other has no github login, generally an email and a user. Manually adding the email fails (there's a unique key on email) and creating the partner fails too (duplicate login). 

<rant>The partner model with its relations to users (and employees) is a total :hankey:</rant>

jgrandguillaume:
Need to use the dedupplicate contact in OCA Odoo instance for this issue

gurneyalex:
@jgrandguillaume can you help here with your deduplicate tool?

jgrandguillaume:
:+1:

gurneyalex:
@StefanRijnhart ok, we made a quick session today with joel to fix the remaining database problems. This script has run, and we should be able to merge and move on with #36

StefanRijnhart:
Great, thanks. Tested but without actually creating users as per my privileges in the OCA database.\

:+1:

guewen:
:+1:

PR #48, opened 2015-02-03
drop testing with python2.6

gurneyalex:
this Python version is really old.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,6 @@ language: python
 
 python:
     - "2.7"
-    - "2.6"
 
 install:
     - pip install coveralls flake8


lmignon:
:+1:

yvaucher:
:+1:

pedrobaeza:
But Odoo still supports it officially, so we shouldn't remove it.

gurneyalex:
@pedrobaeza this is maintainer-tools, i.e. scripts for to help maintainers... Odoo's supported / required Python version has nothing to do with this.

pedrobaeza:
OK then :+1:

PR #51, opened 2015-03-07
Better clone everything

gurneyalex:
* use git fetch on already cloned repos
* try to use the github api to get an up to date list of repositories

### DIFF
--- a/tools/oca_projects.py
+++ b/tools/oca_projects.py
@@ -10,6 +10,8 @@ OCA_REPOSITORY_NAMES: list of OCA repository names
 
 """
 
+from github_login import login
+
 ALL = ['OCA_PROJECTS', 'OCA_REPOSITORY_NAMES', 'url']
 
 OCA_PROJECTS = {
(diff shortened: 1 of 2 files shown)


tools/oca_projects.py
```diff
@@ -10,6 +10,8 @@
 
 """
 
+from github_login import login
```
hbrunn: shouldn't you anticipate an ImportError here?

tools/oca_projects.py
```diff
@@ -107,9 +109,27 @@
...
-OCA_REPOSITORY_NAMES = []
-for repos in OCA_PROJECTS.itervalues():
-    OCA_REPOSITORY_NAMES += repos
+
+def get_repositories():
+    ignored = set([
+        'odoo-community.org',
+        'community-data-files',
+        'contribute-md-template',
+        'website',
```
hbrunn: why?

tools/oca_projects.py
lepistone: Why? `ImportError` is good I think. I would handle specially only error messages which are misleading.

lepistone:
:+1: thanks

tools/oca_projects.py
hbrunn: I understand the following that the script should still work if there's no github login, then it should fall back to the hardcoded list. The same should apply in my opinion if the module is not available

tools/oca_projects.py
mdietrichc2c: I think those OCA repositiories are not needed for every project.

yvaucher: Those OCA repositories are not Odoo modules

yvaucher:
:+1:

tools/oca_projects.py
StefanRijnhart: website and community-data files are.

PR #52, opened 2015-03-13
[ADD] OCA icon

hbrunn:
how about trying to establish a standard icon (so people are welcome so use something more specific, but can use this as fallback)?


rvalyi:
:+1: okay for me

sebastienbeau:
:+1:

max3903:
:+1:

pedrobaeza:
I'm not OK with that, because I don't want to see all modules with the same icon that they are not distinguished. One of the goals of the icons and kanban view is to visually diferentiate each element, and with this, the goal fails.

max3903:
@pedrobaeza This is currently the case today (grey cube icon). The OCA logo is still better than the cube and if you don't like it, you can change it ;)

pedrobaeza:
The grey icon is far more neutral...

rvalyi:
what about a grey OCA icon to adjust Pedro's appeal to monochrome monotony?

lmignon:
@rvalyi  :+1:

pedrobaeza:
Let me make a suggestion in a couple of hours.

rvalyi:
that is fine the OCA will certainly find a consensual color to pass over this merge of death. At the same time it's Python and it's Friday, so I propose a convenient poll model to accommodate to the opinion of the subsequent reviewers:
https://www.youtube.com/watch?v=pWS8Mg-JWSg

pedrobaeza:
I have made #53 with my proposal.

PR #53, opened 2015-03-14
[IMP] Default template icon

pedrobaeza:
Following #52 discussion, I propose a default icon that is more neutral, but distinguishes from the rest of the modules and promotes OCA's brand.


guewen:
Nice!

sebastienbeau:
:+1: 
Good work !

lmignon:
:+1:

rvalyi:
:+1:

max3903:
The logo on the top side would be even nicer.

:+1:

pedrobaeza:
@max3903, I tried to put on all the three visible sides, but it looks weird on the top one. That's why I left it this way.

PR #56, opened 2015-04-13
[ADD] Conventions

max3903:
Move the OCA Conventions from the OCA website


conventions/README.rst
```diff
@@ -0,0 +1,149 @@
...
+====
+
+* Team name must not contain odoo or openerp
+* Team name for localization is "Belgium Maintainers" for Belgium
+
+Project / Repository
+====================
+
+* Project name must not contain odoo or openerp
+* Project name for localization is "l10n_Belgium" for Belgium
```
pedrobaeza: Belgium in lowercase.

dreispt:
Great to have this here! :+1:

hbrunn:
indeed :+1:

yvaucher:
Shouldn't we put this in a CONTRIBUTION.md file ?

https://github.com/blog/1184-contributing-guidelines

CONTRIBUTING.rst
```diff
@@ -0,0 +1,149 @@
...
+Issue
+=====
+
+* Issues are used for blueprints and bugs.
+
+Commit
+======
+
+* Write a helpful commit message
+* Use a commit tag in each message. This tag should be one of:
```
guewen: I don't think we should ask for these "commit tags" in the conventions. It is not anymore widely used since we moved on github. Personally I always preferred to have this kind of information well formed in the sentence itself ("Fixes this", "Refactor that"). When I was using those tags, I observed that it was often difficult to choose one or another (especially between IMP, FIX, REF).

dreispt: @guewen I agree,with you, but I think any change to the current text should be discussed in their own issues or PRs, after this is merged.

guewen: Fine for me!

guewen:
Thanks :+1:

CONTRIBUTING.rst
```diff
@@ -0,0 +1,149 @@
...
+
+* Start by thanking the contributor / developer for his work. No matter the issue of the MP, someone make work for you here, so be thankful for that.
+* Be cordial and polite. Nothing is obvious in a MP.
+* The description of the changes should be clear enough for you to understand his purpose and if apply, contain the reference feature instance in order to allow people to run and test the review
+* Choose the review tag (comment, approve, rejected, needs information,...) and don't forget to add a type of review to let people know:
+
+  * Code review: means you look at the code
+  * Test: means you tested it functionally speaking
+
+While making the merge, please respect the author using the “--author” option when committing. The author is found using the bzr log command. Use the commit message provided by the contributor if any.
```
lmignon: bzr?

PR #58, opened 2015-05-11
Load OCA image via HTTPS

antespi:
This little change avoid HTTPS broken pages in Odoo when user goes to addon details view


dreispt:
:+1:

bwrsandman:
:+1:

PR #60, opened 2015-06-01
initial script to add badges and climate (with pdb)

lepistone:


tools/add-badges.py
```diff
@@ -0,0 +1,56 @@
...
+            sys.stdout.write(readme_line)
+            if not stuff_added and 'travis-ci.org' in readme_line:
+                sys.stdout.write(new_lines)
+        subprocess.check_call(['git', 'add', '--all'])
+        subprocess.check_call(
+            ['git', 'commit', '-m', 'add new badges and codeclimate']
+        )
+
+        # stop here!
+        import pdb; pdb.set_trace()  # XXX BREAKPOINT
```
bwrsandman: Could this be done in two lines? The import can go at the start.

lepistone: Actually I was thinking of getting rid of the debugger once it seems to work on the few ones. But I'm open to advice!

tools/add-badges.py
```diff
@@ -0,0 +1,71 @@
+import re
+import os
+import subprocess
+import fileinput
+import sys
+import shutil
+
+for repo_list_line in open('repositories.txt'):
```
pedrobaeza: Can be `repositories.txt` obtained with https://github.com/OCA/maintainer-tools/blob/master/tools/oca_projects.py? If so, document it here.

tools/add-badges.py
```diff
@@ -0,0 +1,71 @@
...
+        if not os.path.exists('.codeclimate.yml'):
+            new_lines += (
+                "[![Code Climate]"
+                "(https://codeclimate.com/github/OCA/{0}/badges/gpa.svg)]"
+                "(https://codeclimate.com/github/OCA/{0})\n".format(
+                    repo_name
+                )
+            )
+            shutil.copy('../.codeclimate.yml', '.')
+
```
StefanRijnhart: Looks like you are using a static file. Doesn't this file contain a reference to the project?

tools/add-badges.py
```diff
@@ -0,0 +1,71 @@
...
+import sys
+import shutil
+
+for repo_list_line in open('repositories.txt'):
+    m = re.search('(\d+)\|github.com/OCA/(.*)', repo_list_line)
+    repo_name = m.group(2)
+    repo_id = m.group(1)
+    print(repo_name)
+    if not os.path.exists(repo_name):
+        subprocess.call(['hub', 'clone', '--quiet', 'OCA/' + repo_name])
```
pedrobaeza: What is the advantage of hub over git here? If it's only for not writing full URL, I prefer to put it and avoid the dependency.

tools/add-badges.py
```diff
@@ -0,0 +1,71 @@
...
+        try:
+            subprocess.check_call(['git', 'checkout', '--quiet', version])
+        except subprocess.CalledProcessError:
+            continue
+
+        subprocess.check_call(['git', 'reset', '--hard'])
+        subprocess.check_call(['git', 'clean', '-fxd'])
+        new_lines = (
+            "[![Runbot Build Status]"
+            "(http://runbot.odoo-community.org/runbot/badge/flat/{0}/{1}.svg)]"
```
pedrobaeza: All branches are finally served by the same runbot? Or there are 2 URLs?

tools/add-badges.py
```diff
@@ -0,0 +1,71 @@
...
+                "(https://codeclimate.com/github/OCA/{0})\n".format(
+                    repo_name
+                )
+            )
+            shutil.copy('../.codeclimate.yml', '.')
+
+        stuff_added = False
+        for readme_line in fileinput.input('README.md', inplace=1):
+            sys.stdout.write(readme_line)
+            if not stuff_added and 'travis-ci.org' in readme_line:
```
pedrobaeza: `stuff_added` variable doesn't change its value.

tools/add-badges.py
```diff
@@ -0,0 +1,71 @@
...
+
+        subprocess.check_call(['git', 'add', '--all'])
+        subprocess.check_call(
+            ['git', 'commit', '-m', 'add new badges and codeclimate']
+        )
+
+        # stop here!
+        import pdb; pdb.set_trace()  # XXX BREAKPOINT
+
+    os.chdir('..')
```
pedrobaeza: Is this needed?

tools/add-badges.py
```diff
@@ -0,0 +1,71 @@
...
+                transifex_folder
+            )
+        )
+
+        with open("REAMDME.md", "a") as readme_file:  # append
+            readme_file.write(transifex_lines)
+
+        subprocess.check_call(['git', 'add', '--all'])
+        subprocess.check_call(
+            ['git', 'commit', '-m', 'add new badges and codeclimate']
```
pedrobaeza: Put a complete commit message like: `[IMP] Add new badges: runbot, codeclimate and Transifex`

pedrobaeza: Put as author OCA bot or the user for these purposes.

tools/add-badges.py
pedrobaeza: And where is that file? It's not in this PR

tools/add-badges.py
bwrsandman: One URL for the build status (svg) and one for the linked page which had all the branches and PRs

pedrobaeza: OK then

tools/add-badges.py
bwrsandman: This file should be added to the sample_files in MQT
```yaml
languages:
  JavaScript: true
  Python: true
exclude_paths:
- "__unported__/*"
- "*/__openerp__.py"
- "*/migrations/*"
```

tools/add-badges.py
lepistone: yes

tools/add-badges.py
lepistone: done: OCA/maintainer-quality-tools#201

tools/add-badges.py
lepistone: Also, it figures out the correct URL based on your rights and configuration (ssh, https, etc). This is a maintainer tool we'll use few times, very carefully. I don't see a problem with a little open-source dependency here.

tools/add-badges.py
lepistone: well spotted! fixed

tools/add-badges.py
lepistone: I am not sure about this: the author of the commit is still me, or whoever will run this. Useful to blame someone in case of trouble, like for manual commits.

In any case, I expanded the commit message to say that it is an automated commit.

lepistone:
I answered to the feedback, thanks!

tools/add-badges.py
pedrobaeza: OK, don't worry then about this.

tools/add-badges.py
```diff
@@ -0,0 +1,85 @@
...
+                transifex_folder
+            )
+        )
+
+        with open("REAMDME.md", "a") as readme_file:  # append
+            readme_file.write(transifex_lines)
+
+        subprocess.check_call(['git', 'add', '--all'])
+        subprocess.check_call(
+            ['git', 'commit', '-m',
```
pedrobaeza: If you put `--author`, you still appears as committer, but not author. I prefer this formula to easy the search of these kind of commits accross repositories.

tools/add-badges.py
pedrobaeza: OK, I see that you enter each directory with a relative path.

StefanRijnhart:
Thanks! :+1: (no test)

tools/add-badges.py
bwrsandman: I agree

lepistone: OK. what should the author be exactly?

bwrsandman:
:+1:

tools/add-badges.py
```diff
@@ -0,0 +1,85 @@
...
+             '''\
+add badges and files for transifex, runbot, codeclimate
+
+This is an automatic commit done with the add-badges.py script
+in maintainers-tools.
+''']
+        )
+
+        # stop here!
+        import pdb; pdb.set_trace()  # XXX BREAKPOINT
```
bwrsandman: Not pep8, add the import at the top of the script

lepistone: done

PR #62, opened 2015-06-14
Update CONTRIBUTING.md

samuellefever:
Remove duplicated text

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -184,13 +184,6 @@ Use the following pattern:
 </record>
 ```
 
-These rules should be followed even when a view is inherited, the module name
-prevents xid conflicts. In the case where a view inherited has a name which
-does not follow the guidelines set above, prefer naming the inherited view
-after the original over using a name which follows the guidelines. This eases
-looking up the original view and other inheritance if they all have the same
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -184,13 +184,6 @@ Use the following pattern:
 </record>
 ```
 
-These rules should be followed even when a view is inherited, the module name
-prevents xid conflicts. In the case where a view inherited has a name which
-does not follow the guidelines set above, prefer naming the inherited view
-after the original over using a name which follows the guidelines. This eases
-looking up the original view and other inheritance if they all have the same
-name.
-
```
dreispt: I'm sorry, I don't see the duplication.

yvaucher:
:-1: I don't see the duplication either

CONTRIBUTING.md
guewen: @dreispt look at the end of this section https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#security-view-and-action and at the beginning of the next one

guewen:
:+1:

yvaucher:
:+1:

PR #63, opened 2015-06-16
Start Transifex sync script

samuellefever:
I started the script using github3 library.
It gets the translations, compares it with what is in the repository, and do the commit only if needed.

I am blocked with a error on line 88 when I try to commit.
```
github3.exceptions.UnprocessableEntity: 422 Invalid request.
"email", "name" weren't supplied.
```

Any help is welcome.

### DIFF
--- a/tools/tx_pull.py
+++ b/tools/tx_pull.py
@@ -1,17 +1,18 @@
 # -*- coding: utf-8 -*-
 
 """
-This script is meant to pull the translations from Transilex .
-Technically, it will pull the translations from transilex,
+This script is meant to pull the translations from Transifex .
+Technically, it will pull the translations from transifex,
 compare it with the po files in the repository and replace it if needed
 """
 
 from transifex.api import TransifexAPI
-import os, shutil
+import os
+import shutil
 import difflib
 import re
-from config import read_config, write_config
-from oca_projects import OCA_REPOSITORY_NAMES
+from config import read_config
+# from oca_projects import OCA_REPOSITORY_NAMES
 
 config = read_config()
 # Github parameter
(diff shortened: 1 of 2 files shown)


max3903:
You need to provide an identity to commit. Use "Transbot <transbot@odoo-community.org>".

tools/tx_pull.py
```diff
@@ -0,0 +1,127 @@
+# -*- coding: utf-8 -*-
+
+"""
+This script is meant to pull the translations from Transiflex .
+Technically, it will pull the translations from transiflex,
```
max3903: Transifex, no 'l'

yvaucher:
@samuellefever Isn't it your git author configuration that is missing an email? 
git config --global user.email "email"

tools/tx_pull.py
samuellefever: I tries with author = {'name' : 'samuellefever', 'email' : 'sam@niboo.be'} and it stays the same

samuellefever:
As I am not using my local git config but the github3 library, it did not fix the problem

max3903:
If I am not mistaken, @pedrobaeza told me today that he will test this script.

bwrsandman:
@pedrobaeza how goes the test?

tools/tx_pull.py
```diff
@@ -0,0 +1,127 @@
+# -*- coding: utf-8 -*-
+
+"""
+This script is meant to pull the translations from Transifex .
+Technically, it will pull the translations from transifex,
+compare it with the po files in the repository and replace it if needed
+"""
```
max3903: Could you please document here how to install its dependencies (Transifex), how to configure it and how use it ?

pedrobaeza:
Superseeded by #69

PR #66, opened 2015-06-29
fix-exception UserError not in v8

bealdav:
The path to go from v8 to v9 is soft with this rule: 
- only import change
- inside use of UserError of the same for both


guewen:
:+1:

pedrobaeza:
Good trick with future version in mind :+1:

jgrandguillaume:
:+1:

bealdav:
@pedrobaeza it's an idea of @sebastienbeau. Surely a good idea

pedrobaeza:
You all guys have very good ideas, and you also share with community, so it's awesome!

PR #78, opened 2015-07-13
make package pip installable and importable

charbeljc:
- move `tools` to `oca.tools`
- rename `clone_everything` to `clone_everything.py`
- make it importable without network connection.


guewen:
>    move tools to oca.tools

Ok

>    rename clone_everything to clone_everything.py

Yes, so it can be called from the entry point (or from other scripts).
Could you do the same for `check_contrib`?


>    make it importable without network connection.

I would have preferred if the constant `OCA_REPOSITORY_NAMES` was deprecated and replaced by a function but that's still better now.

charbeljc:
@guewen done for `check_contrib` -> `check_contrib.py`

guewen:
Thanks :+1:

bwrsandman:
:+1:

setup.cfg
```diff
@@ -15,13 +15,13 @@ classifier =
     Programming Language :: Python
 [files]
 packages =
-    tools
+    oca
```
pedrobaeza: Shouldn't it be oca-maintainer-tools to allow other future pip packages (like oca-maintainer-quality-tools)?

lmignon:
:-1: You should define the module as part of a namespace package in the setup.py file so it will be possible to release others OCA modules under the same root package. (oca)
I'm in favor to make this module 'pip installable'  but we must take care of the way packages are named.
IMO a proper way to distribute these tools as part of of python module can be done by naming the module as 'oca.tools' and define it as a namespace package (https://pythonhosted.org/setuptools/setuptools.html#namespace-packages). 
```python
setuptools.setup(name='oca.tools',
                 description="OCA maintainer tools",
                 ...
                 namespace_packages=['oca', 'oca.tools']
                 ...
)
                
``` 
`oca/__init__.py` and `oca/tools/__init__.py`  must contain the line 
```python
# See http://peak.telecommunity.com/DevCenter/setuptools#namespace-packages
try:
    __import__('pkg_resources').declare_namespace(__name__)
except ImportError:
    from pkgutil import extend_path
    __path__ = extend_path(__path__, __name__)
```
This code ensures that the namespace package machinery is operating and that the current package is registered as a namespace package.

By packaging the module in this way, it remains possible to also distribute the quality tools under the same root package 'oca'.

charbeljc:
Hi all, apologizes for the late feedback, trying to feed Odoo with a moderate sized list of partners can be quite a disappointing experience.

@lmignon thanks for the namespace package machinery trick, I will redo this PR as soon I am ok with the next step (inter partner relationship).

Regards

nhomar:
Hello.

My comment can be done here or can be done in a separat one.

If you use a good docstring with examples you will be able to jump the namespace trick and use the doc examples for testing purpose.

How? using the doctest module on python.

Here some examples:

https://docs.python.org/2/library/doctest.html

For me the scope of this PR is covered, we can finish such job in a different PR.

then +1 for me,.

nhomar:
you need to rebase it.

setup.cfg
moylop260: Im agree with @pedrobaeza 
Maybe, oca_mt
To work in a future next package with oca_* (maybe oca_mqt to check lints)

nhomar:
@charbeljc take care about name chosen, on pip "oca" exist. https://pypi.python.org/pypi/oca/0.2.3

lmignon:
@nhomar do you have read my comment https://github.com/OCA/maintainer-tools/pull/78#issuecomment-121167522? We need to fix the setup.py to properly define a reusable namespace for all oca modules. It's a little change to do and I hope it will be done before merging this PR.

nhomar:
Yes I read it! I am just mentioning the point to bring alive some activity, conceptuiually the PR is fine and the namespace stuf I am agreed,

My +1 comes after changes asked comes.

Regards.

lmignon:
@nhomar Thank you for the clarification.

charbeljc:
Hi @nohmar and @lmignon, I finally took some time to advance this PR, starting implementing your namespace machinery. I need your assistance on this point, or maybe I overlooked something.

I first inserted namespace declaration in both oca/__init__.py and oca/tools/__init__.py, but with the last addition, I loose the ability to declare package content with __all__, which is a good thing when you are exploring modules under ipython or other tools. As the doc says, when you put a namespace declaration in your __init__.py, you cannot put anything else in it.

Any advice on this point ?

I would like finalize this PR today.

Regards

Re

lmignon:
@charbeljc I've fixed the remaining issues and submitted a PR on your branch https://github.com/charbeljc/maintainer-tools/pull/1

setup.py
```diff
@@ -1,6 +1,8 @@
 import setuptools
 
-setuptools.setup(setup_requires=['pbr'],
+requires = open('requirements.txt', 'r').read().split('/n')
```
lepistone: do we actually need `pbr`?

oca/__init__.py
```diff
@@ -0,0 +1,7 @@
+# -*- coding: utf-8 -*-
+# See http://peak.telecommunity.com/DevCenter/setuptools#namespace-packages
+try:
+        __import__('pkg_resources').declare_namespace(__name__)
```
lepistone: indent

oca/tools/__init__.py
```diff
@@ -0,0 +1,7 @@
+# -*- coding: utf-8 -*-
+# See http://peak.telecommunity.com/DevCenter/setuptools#namespace-packages
+try:
+        __import__('pkg_resources').declare_namespace(__name__)
```
lepistone: indent

lepistone:
Hi, I made a few minor remarks. Is the fact that `oca` is taken in pypi resolved? Shouldn't we take another namespace?

Thanks for your work, probably great, but I'm not confident enough with setuptools to approve.

lepistone:
Anyway, this needs a rebase. Thanks!

charbeljc:
@lepistone, thx for your review, rebase done.

As for the name change, since `oca` is not available on pypi,  I'm open to suggestions. @moylop260 suggested 'oca_mt', what about `otools` ? Same number of `o` as in `Odoo`, sounds a bit like Laurence from Arabia Peter O'Toole ...

As a side question, why do we make a distinction between `maintainer-tools` and `maintainer-quality-tools` ?

lepistone:
`o-toole`! Great!

dreispt:
MQT is about Continuous Integration: running  automated testes a style checks.
MT is an utility belt for OCA developers and admins.

charbeljc:
Hi @dreispt, I still can't see the rationale for such separation ...

osvalr:
a rebase is needed here

gurneyalex:
old and author is no longer working on odoo.

PR #79, opened 2015-07-13
[ADD] README.rst generator from __openerp__.py

charbeljc:
* add a requirement on https://github.com/charbeljc/pystache.git, waiting for the following pending PRs:
   * https://github.com/defunkt/pystache/pull/176
   * https://github.com/defunkt/pystache/pull/177
   * https://github.com/defunkt/pystache/pull/178

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 # Config file .travis.yml
+sudo: false
 
 language: python
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,4 +5,5 @@ bzr==2.6.0
 github3.py==0.9.0
 requests==2.3.0
 uritemplate.py==0.3.0
+-e git+https://github.com/charbeljc/pystache.git#egg=pystache-master
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -25,3 +25,4 @@ console_scripts =
     oca-set-repo-labels = tools.set_repo_labels:main
     oca-odoo-login = tools.odoo_login:main
     oca-sync-users = tools.oca_sync_users:main
+    oca-make-readme = tools.modoornize:main
--- a/template/module/README.rst.mustache
+++ b/template/module/README.rst.mustache
@@ -1,0 +1,43 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+{{= name}}
+
+{{description}}
+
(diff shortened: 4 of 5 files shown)


dreispt:
Why add another templating lib when we already have so many options available? 
I believe that the same effect could be easilly achieved with `.format()`...

charbeljc:
Hi @dreispt, thank you for your comments. Maybe I would have achieved the same effect using `.format()`, but with less elegance. With my additions to `pystache`, you can for instance write things like this.

```python
...
def under(a):
   return '\n'.join([a, '=' * len(a)])
renderer.register_helper('=', under) 
print renderer.render('...\n{{= name}}\n...', {'name': 'Section 1'})
``` 
This will expand to:
```
...
Section 1
=========
...
````
Besides, `pystache` as many other useful features making it a valuable addition in our toolchain.

Regards

dreispt:
Personally, I try to avoid additional dependencies as much as I can. But that's my personal opinion.

guewen:
I tend to agree with @dreispt and if we have to choose to add a dependency, I would consider other choices, such as for instance cookiecutter which can also injects user inputs and default values (in user's configuration files) (which we'll need for filling the templates).

moylop260:
@charbeljc 
Good job, good feature!
:+1:

moylop260:
@guewen and @dreispt 
I'm agree with you.
It can use webkit template too.
However, we need to add new library to requirements.txt too.
Maybe `pystache` is a good library too (I don't know)
We will need to talk about what is the better library to use templates in OCA. (other issue).
But now, If we don't have this rule IMHO we can't to block PR, what do you think?

dreispt:
We already have Jinja and Mako as Odoo dependencies. I honestly don't see the need for yet  another templating lib.

moylop260:
We can define then that our templates script should be based on Mako or Jinja.
Right?

nhomar:
I this time prefer do this based on cookiecutter which use jinja BTW and is really good tested.

We know Mako, Jinja now we Learn Qweb, Honestly I do not have any preference, but I do not want to use another one.

Regards.

charbeljc:
Hum, @guewen, @nhomar `cookiecutters` looks nice, I will test it later, as soon as my projects deadlines are done, thanks for the pointer

max3903:
I agree with @guewen and @dreispt 

:-1:

gurneyalex:
:-1:

closing.

PR #88, opened 2015-07-24
Add guidelines for placing reports

### LINKED ISSUE #87: No guidelines given for placing reports in modules
The guidelines state where we should store the models, views etc, but does not indicate where reports must go.

clonedagain:
Seems to be the current convention used at Odoo SA's.
Fixes #87

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -21,14 +21,14 @@ section.
 
 A module is organised in a few directory:
 
+* `controllers/`: contains controllers (http routes)
 * `data/`: data xml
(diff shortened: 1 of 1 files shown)


dreispt:
:+1:

dreispt:
Should that list have a criteria to order it's entries?

clonedagain:
Updated PR to indicate where to place the report files in the complete tree. Is it correct?

guewen:
Follows the changes in Odoo, :+1: 
Thanks!

bwrsandman:
:+1:

PR #89, opened 2015-07-24
Fix guidelines for lambda default

clonedagain:
I think the proposed lambda should call the default method, not return it.


dreispt:
:+1:

guewen:
:+1:

PR #90, opened 2015-07-26
Add guideline for version numbers

dreispt:
Rationale:

It is unfortunate that most modules don't include in the manifest the information of the major Odoo version they target. This is an important information since most of the time modules are incompatible with other Odoo version other than the one they target.
For example: it is frequent to find code in Github with an Odoo module in a "master" branch, but can't be sure of what Odoo version it targets by looking at it's manifest file.

This is so because in the past Odoo/Openerp automatically added the major version to module versions. Thus if your module was version "6.1.1", it would be displayed as "6.1.6.1.1".

It is no longer the case: [now](https://github.com/odoo/odoo/blob/8.0/openerp/modules/module.py#L368) Odoo does not do that if the major version is already appended to the module version!

Thus the Odoo target version can be included in the module version without being duplicated by the loading functions, and properly document the module's Odoo target version.

PR #91, opened 2015-07-27
Add template magic files. Required in all modules.

Yajo:
Add missing files `__openerp__.py` and `__init__.py`.

### DIFF
--- a/template/module/__init__.py
+++ b/template/module/__init__.py
@@ -1,0 +1,1 @@
+# -*- encoding: utf-8 -*-
--- a/template/module/__init__.py
+++ b/template/module/__init__.py
@@ -1,1 +1,1 @@
-# -*- encoding: utf-8 -*-
+# -*- coding: utf-8 -*-
--- a/template/module/__init__.py
+++ b/template/module/__init__.py
@@ -1,1 +1,3 @@
 # -*- coding: utf-8 -*-
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+# © <YEAR(s)> <AUTHOR(s)>
--- a/template/module/__init__.py
+++ b/template/module/__init__.py
@@ -1,3 +1,3 @@
 # -*- coding: utf-8 -*-
+# © <YEAR(S)> <AUTHOR(S)>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
-# © <YEAR(s)> <AUTHOR(s)>
--- a/template/module/__openerp__.py
+++ b/template/module/__openerp__.py
@@ -6,9 +6,9 @@
     "summary": "Module summary",
     "version": "8.0.1.0.0",
     "category": "Uncategorized",
-    "license": "AGPL-3",
     "website": "https://odoo-community.org/",
     "author": "<AUTHOR(S)>, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
     "application": False,
     "installable": True,
     "external_dependencies": {
--- a/template/module/__openerp__.py
+++ b/template/module/__openerp__.py
@@ -1,18 +1,7 @@
 # -*- coding: utf-8 -*-
-# Odoo, Open Source Management Solution
-#
-# This program is free software: you can redistribute it and/or modify
-# it under the terms of the GNU Affero General Public License as published by
-# the Free Software Foundation, either version 3 of the License, or
-# (at your option) any later version.
-#
-# This program is distributed in the hope that it will be useful,
-# but WITHOUT ANY WARRANTY; without even the implied warranty of
-# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-# GNU Affero General Public License for more details.
-#
-# You should have received a copy of the GNU Affero General Public License
-# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+# © <YEAR(s)> <AUTHOR(s)>
+
 # Instructions: https://www.odoo.com/documentation/8.0/reference/module.html
 {
     "name": "Module name",
--- a/template/module/__openerp__.py
+++ b/template/module/__openerp__.py
@@ -2,14 +2,25 @@
 {
     "name": "Module name",
     "summary": "Module summary",
-    "version": "0.1.0",
-    "author": "Odoo Community Association (OCA)",
+    "version": "8.0.1.0.0",
     "category": "Module category",
+    "license": "AGPL-3",
+
+    "website": "https://odoo-community.org/",
+    "author": "Odoo Community Association (OCA)",
+
+    "application": True,
+    "installable": True,
+
+    "external_dependencies": {
+        "python": [],
+        "bin": [],
+    },
+
     "depends": [
-        "base",
-    ],
-    "demo": [
     ],
     "data": [
+    ],
+    "demo": [
     ],
 }
--- a/template/module/__openerp__.py
+++ b/template/module/__openerp__.py
@@ -5,18 +5,14 @@
     "version": "8.0.1.0.0",
     "category": "Module category",
     "license": "AGPL-3",
-
     "website": "https://odoo-community.org/",
     "author": "Odoo Community Association (OCA)",
-
     "application": True,
     "installable": True,
-
     "external_dependencies": {
         "python": [],
         "bin": [],
     },
-
     "depends": [
     ],
     "data": [
--- a/template/module/__openerp__.py
+++ b/template/module/__openerp__.py
@@ -1,4 +1,4 @@
-# -*- encoding: utf-8 -*-
+# -*- coding: utf-8 -*-
 # Instructions: https://www.odoo.com/documentation/8.0/reference/module.html
 {
     "name": "Module name",
--- a/template/module/__openerp__.py
+++ b/template/module/__openerp__.py
@@ -1,4 +1,5 @@
 # -*- encoding: utf-8 -*-
+# Instructions: https://www.odoo.com/documentation/8.0/reference/module.html
 {
     "name": "Module name",
     "summary": "Module summary",
--- a/template/module/__openerp__.py
+++ b/template/module/__openerp__.py
@@ -1,8 +1,6 @@
 # -*- coding: utf-8 -*-
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 # © <YEAR(s)> <AUTHOR(s)>
-
-# Instructions: https://www.odoo.com/documentation/8.0/reference/module.html
 {
     "name": "Module name",
     "summary": "Module summary",
--- a/template/module/__openerp__.py
+++ b/template/module/__openerp__.py
@@ -5,7 +5,7 @@
     "name": "Module name",
     "summary": "Module summary",
     "version": "8.0.1.0.0",
-    "category": "Module category",
+    "category": "Uncategorized",
     "license": "AGPL-3",
     "website": "https://odoo-community.org/",
     "author": "Odoo Community Association (OCA)",
--- a/template/module/__openerp__.py
+++ b/template/module/__openerp__.py
@@ -1,4 +1,18 @@
 # -*- coding: utf-8 -*-
+# Odoo, Open Source Management Solution
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
 # Instructions: https://www.odoo.com/documentation/8.0/reference/module.html
 {
     "name": "Module name",
--- a/template/module/__openerp__.py
+++ b/template/module/__openerp__.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
+# © <YEAR(S)> <AUTHOR(S)>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
-# © <YEAR(s)> <AUTHOR(s)>
 {
     "name": "Module name",
     "summary": "Module summary",
--- a/template/module/__openerp__.py
+++ b/template/module/__openerp__.py
@@ -8,7 +8,7 @@
     "license": "AGPL-3",
     "website": "https://odoo-community.org/",
     "author": "Odoo Community Association (OCA)",
-    "application": True,
+    "application": False,
     "installable": True,
     "external_dependencies": {
         "python": [],
--- a/template/module/__openerp__.py
+++ b/template/module/__openerp__.py
@@ -8,7 +8,7 @@
     "category": "Uncategorized",
     "license": "AGPL-3",
     "website": "https://odoo-community.org/",
-    "author": "Odoo Community Association (OCA)",
+    "author": "<AUTHOR(S)>, Odoo Community Association (OCA)",
     "application": False,
     "installable": True,
     "external_dependencies": {
--- a/template/module/__openerp__.py
+++ b/template/module/__openerp__.py
@@ -1,0 +1,15 @@
+# -*- encoding: utf-8 -*-
+{
+    "name": "Module name",
+    "summary": "Module summary",
+    "version": "0.1.0",
+    "author": "Odoo Community Association (OCA)",
+    "category": "Module category",
+    "depends": [
+        "base",
+    ],
+    "demo": [
+    ],
+    "data": [
+    ],
+}
--- a/template/module/__openerp__.py
+++ b/template/module/__openerp__.py
@@ -16,6 +16,7 @@
         "bin": [],
     },
     "depends": [
+        "base",
     ],
     "data": [
     ],


template/module/__openerp__.py
```diff
@@ -0,0 +1,15 @@
+# -*- encoding: utf-8 -*-
+{
+    "name": "Module name",
+    "summary": "Module summary",
+    "version": "0.1.0",
```
dreispt: Based on the arguments in #90,  suggest using `8.0.1.0` as the default version.

template/module/__openerp__.py
```diff
@@ -0,0 +1,15 @@
...
+    "author": "Odoo Community Association (OCA)",
+    "category": "Module category",
+    "depends": [
+        "base",
+    ],
+    "demo": [
+    ],
+    "data": [
+    ],
```
dreispt: Missing:

    "installable": True,
    "license": "AGPL-3",

template/module/__openerp__.py
```diff
@@ -0,0 +1,15 @@
+# -*- encoding: utf-8 -*-
+{
+    "name": "Module name",
+    "summary": "Module summary",
+    "version": "0.1.0",
+    "author": "Odoo Community Association (OCA)",
+    "category": "Module category",
+    "depends": [
+        "base",
+    ],
```
bwrsandman: A little known key is `external_dependencies`. I would argue that adding it in the template would help newer devs who do not know about it yet.
```
    'external_dependencies': {
        'python': [],
        'bin': [],
    },
```

template/module/__openerp__.py
```diff
@@ -0,0 +1,15 @@
+# -*- encoding: utf-8 -*-
+{
+    "name": "Module name",
+    "summary": "Module summary",
+    "version": "0.1.0",
+    "author": "Odoo Community Association (OCA)",
+    "category": "Module category",
+    "depends": [
+        "base",
```
bwrsandman: "base" is redundant

template/module/__openerp__.py
bwrsandman: * `"application"`
* `"qweb"`

bwrsandman:
Since this is a template, it could be a good place to inline document what each key does.

For example, we can say that specifying a category creates an entry in `ir_module_category`

template/module/__openerp__.py
yvaucher: Following my comment on #90 I would prefer `8.0.1.0.0`

template/module/__openerp__.py
yvaucher: What about `"website"`?

Should it link to the github page of the repo?

bwrsandman: I would say the OCA page of the repo from https://odoo-community.org/page/List

yvaucher: @bwrsandman I like your idea. Thus you mean the team which gather multiple repo? For exemple https://odoo-community.org/project/accounting-maintainers-9
Because we only direct to github page then.

bwrsandman: I guess so.

Yajo: Isn't `qweb` deprecated? Qweb now goes into data files AFAIK.

Yajo:
It could be a good idea, but personally I hate deleting docs from templates and prefer to have them separated. If there are good docs anywhere (not [this](https://www.odoo.com/documentation/8.0/reference/module.html), please), a simple one-line comment in the template like *Read docs at http://example.com* should be enough.

dreispt:
@bwrsandman I agree with @Yajo

bwrsandman:
@Yajo :+1: for a single line pointing to doc

template/module/__openerp__.py
```diff
@@ -0,0 +1,26 @@
+# -*- encoding: utf-8 -*-
+{
+    "name": "Module name",
+    "summary": "Module summary",
+    "version": "8.0.1.0.0",
+    "category": "Module category",
+    "license": "AGPL-3",
+
```
pedrobaeza: Don't put extra empty lines

pedrobaeza:
https://www.odoo.com/documentation/8.0/reference/module.html is not a bad link for reference, but maybe we can improve the documentation. We can make a PR against the documentation branch for that.

Yajo:
@pedrobaeza Seems the best option.

template/module/__openerp__.py
```diff
@@ -0,0 +1,23 @@
+# -*- encoding: utf-8 -*-
```
dreispt: s/encoding/coding

template/module/__init__.py
```diff
@@ -0,0 +1 @@
+# -*- encoding: utf-8 -*-
```
dreispt: s/encoding/coding

template/module/__openerp__.py
Yajo: That's one of the reasons I need this in the template. I always confuse those two.

template/module/__openerp__.py
```diff
@@ -0,0 +1,23 @@
...
+# Instructions: https://www.odoo.com/documentation/8.0/reference/module.html
+{
+    "name": "Module name",
+    "summary": "Module summary",
+    "version": "8.0.1.0.0",
+    "category": "Module category",
+    "license": "AGPL-3",
+    "website": "https://odoo-community.org/",
+    "author": "Odoo Community Association (OCA)",
+    "application": True,
```
dreispt: Do we want this to be the default? I would prefer False. 
Opinions?

Yajo: Hmm I think me too.

yvaucher:
What about the license? Not needed any more in `__openerp__.py` ? We need clear guidance about this too, if we don't put it in this template people won't add it.

I am for not having the license in the `__init__.py` as it posses no logic only imports. We could also remove it from `__openerp__.py` relying on `"license"` key. But in that case we should ask to add a LICENSE file and/or put license header in all source files.

For now we have in all our modules the license in `__openerp__.py` so I would prefer to have as default the AGPL in this template.

pedrobaeza:
I agree with @yvaucher in the __openerp__.py part.

Yajo:
You might want to read [this](http://hackerboss.com/get-rid-of-templates/). IANAL, but Odoo S.A. [have removed all legal boilerplate in master](https://github.com/odoo/odoo/commit/0bd454534818941bc3216025ebe3dcac3d78c19d), and they have asked good lawyers before that for sure. As always, just my opinion.

dreispt:
Individual code files can be separated from their module and repo.
Thus just redirecting licensing info to some other file is not enough.
You should at least have a short notice identifying authorship and the license, but you can leave to long details to another file (OSI licenses full text can be easily found in the web, anyway).

Yajo:
Then why Odoo removed it? (Just curiosity)

yvaucher:
@Yajo I would prefer to refer to this than from a random blogpost:

http://www.gnu.org/licenses/gpl-faq.html#LicenseCopyOnly
http://www.gnu.org/licenses/gpl-faq.html#NoticeInSourceFile

http://www.gnu.org/licenses/agpl-3.0.html
"""
How to Apply These Terms to Your New Programs
To do so, attach the following notices to the program.
It is safest to attach them to the start of each source file to most effectively state the exclusion of warranty;
**and each file should have at least the "copyright" line and a pointer to where the full notice is found.**
"""

So if we prefer, and it seems we do, a notice pointing at a LICENSE file it is possible but we need to add the notice and the license file in templates. Last time we talked about it there was some people in favour of a smaller header with a link to a LICENSE file. However as it was on time of Odoo looking for a new license we took no action as we didn't wanted to make big changes on license.

A notice on a licence (even a one line) is necessary in all sources files as each file can be downloaded individually and a software can include multiple files with multiple licenses

I you want to dig out this talk:
https://lists.launchpad.net/openerp-community/msg07097.html

nhomar:
@yvaucher agreed with you about the option on 1 liner.

Yajo:
What about this?:

```
# -*- coding: utf-8 -*-
# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
# Copyright 2015: See README.rst and Git history.
```

pedrobaeza:
I have started to use this one that allows to scale when more contributors are added: https://github.com/OCA/social/pull/2/files#diff-72b773ba9b4b0c2bb037881c4f9a27c7R3. I underestimate to put the author information in each file in the past, but it's important to be put.

Yajo:
Those `##############################` are very ugly. They give much visual importance to the least important piece of the file.

Besides that, what about this then?:

```
# -*- coding: utf-8 -*-
# License AGPL-3.0 or later. See LICENSE file in root folder for details.
# Copyright 2015: See README.rst and Git history.
```

Yajo:
BTW, XML files still have no license nor copyright disclaimers at all, and nobody cares...

pedrobaeza:
I don't see too bad the  `##############################`, but they can be removed. It's not the main question. XML files are less sensitive to copyright protection I think.

Yajo:
It's a common concept because XML are data files. However, in Odoo XML are the views and they should be considered code. On the other hand, `__openerp__.py` is just a manifest with metadata. It could be a JSON file. I don't consider it code, but most people do just because it ends with `.py`.

FWIW, see what our lawyer-powered neighbors are doing:

- [Odoo S.A. removed all headers and unified all copyrights in one file](https://github.com/odoo/odoo/commit/0bd454534818941bc3216025ebe3dcac3d78c19d).

- The Linux kernel is one of the most successful GPL projects and does not seem to care very much in putting the header of license or copyright in all files ([example](https://github.com/torvalds/linux/blob/master/block/blk-merge.c), [example](https://github.com/torvalds/linux/blob/master/kernel/bounds.c), [example](https://github.com/torvalds/linux/blob/master/kernel/Makefile)). Google, Red Hat and Oracle colud have made their own private forks, but nobody has legally ever done that AFAIK.

- The OCA itself does not care at all about XML copyright or licenses, altough they contain logic in many cases (what's the difference between Qweb's `<t t-if="something">` and Python's `if something:`?).

- Python itself [has its own license](https://hg.python.org/cpython/file/643b332465d7/LICENSE) and [seems to follow the trend](https://hg.python.org/cpython/file/643b332465d7/Python/strdup.c).

- [GitHub says](https://help.github.com/articles/open-source-licensing/#what-happens-if-i-dont-choose-a-license):

  > You're under no obligation to choose a license. It's your right not to include one with your code or project, but please be aware of the implications. Generally speaking, the absence of a license means that **the default copyright laws** apply. This **means that you retain all rights to your source code** and that nobody else may reproduce, distribute, or create derivative works from your work.

On the other side:

- Apple's WebKit has their BSD stuff [in every file I checked](https://svn.webkit.org/repository/webkit/trunk/Source/WebKit2/UIProcess/TextChecker.h).

My vote goes for removing everything, as we are covered in the worst case by what GitHub said. But if you prefer to have a minimal header (it has some logic, after all), I prefer https://github.com/OCA/maintainer-tools/pull/91#issuecomment-127913513 above @pedrobaeza's because it does not need maintenance. If the license can be referenced, then why not do the same with the copyright?

yvaucher:
To be able to put a short notice we must first add LICENSE file in each module (when downloading a module from app.odoo.com you only get the modules not the whole repository).

Yajo:
Hmm I did not know. We could stick with https://github.com/OCA/maintainer-tools/pull/91#issuecomment-127898673.

dreispt:
If you look at Odoo sources, you will find bundled libs from other authors, and if were not the headers in the files, we wouldn't know where to get more information on the library (docs, authors, updates, ...).

I like Pedro's version without the `#####` lines: two lines where you state the License, Year, and Author(s). 
I find that adding a third line with the project's hosting address would be good for practical reasons. 

Something like:

    # (c) 2015 Pedro M. Baeza
    # License AGPL-3 (http://www.gnu.org/licenses/agpl.html)
    # More information at: https://github.com/OCA/maintainer-tools

Yajo:
> `# More information at: https://github.com/OCA/maintainer-tools`

That's already found in the README, which is easily visible in code, in single-module downloads and in GUI. That's also another line to maintain (must change it in each repo), and adds nothing to the copyrighting/licensing matter, so I see no need for it.

Yajo:
Most of you seem to agree in reducing the header, so I updated it.

I added it to `__init__.py` because usually modules will put some code there. Also, if everyone copies from `__openerp__.py`, it could make them copy the instructions line, which is not desired. If not, it does not disturb anyway, and if you do not want it, removing is always easier than adding.

https://github.com/OCA/maintainer-tools/pull/91#issuecomment-127580900 seems the most sensible. I put 2 lines according to the minimal requirements stated by the FSF.

The license is referenced from the FSF website to avoid https://github.com/OCA/maintainer-tools/pull/91#issuecomment-127942240 problems.

The license says "or a later version" because this has been the case until now, and that must be stated too.

I have used © instead of Copyright or (C) because it's smaller (1 char) and BTW we force anyone to use UTF-8 (otherwise they will see strange characters).

pedrobaeza:
Thanks for taking care of this topic.

:+1: to this header.

yvaucher:
About pointer to license I would rather have it to a local LICENSE file in each module. This might sounds dumb but if we want to have 1 repo = 1 module for version 9.0 it would help. Plus it would be downloaded via apps.odoo.com. And of course there is the "gnu.org" could be unreachable issue (for various reasons).

Yajo:
> have 1 repo = 1 module for version 9.0

Will that happen?

pedrobaeza:
> have 1 repo = 1 module for version 9.0
> Will that happen?
For now it's very unlikely

> About pointer to license I would rather have it to a local LICENSE file in each module
@yvaucher I don't think necessary to have this overload on each module. A link is enough to justify a license, and a temporary unreachable issue doesn't prevent the terms of the license.

yvaucher:
@Yajo I'd like that to happen and I'm not the only one. But it's not my call. There was the same discution over version numbers.

pedrobaeza:
I have to clear that I also want it, but we have a lot of technical troubles for making it. We can discuss this in another issue/thread.

Yajo:
Then that seems out of scope for now.

dreispt:
As I think Pedro mentioned before, modules downloaded from Odoo Apps are distributed isolated from the repo. If only the repo has a license file, it will be missing from the module.
Also, in the same repo we can find modules with different linceses.

For these two reasons, and if linking to the GNU site is not enough, then each module should dundle it's full license text.

Yajo:
The last paragraph of the official FSF recommendation reads:

> You should have received a copy of the GNU Affero General Public License along with this program.  If not, see <http://www.gnu.org/licenses/>.

Which means that **not receiving that file does not invalidate the license**, since it can be accessed from that URL that is clearly stated in the file header. With our current header, we do basically the same with less words.

yvaucher:
@Yajo 
http://www.gnu.org/licenses/gpl-howto.html

    [...]
    You should also include a copy of the license itself somewhere in the distribution of your program.
    All programs, whether they are released under the GPL or LGPL, should include the text
    version of the GPL. In GNU programs the license is usually in a file called COPYING.
    [...]

apps.odoo.com is a way of distributing. Plus, multiple modules in a single repository could have different licenses. One single license file at repo root doesn't help with multiple licenses.

The file http://www.gnu.org/licenses/agpl.txt is only 34 kB and wouldn't be that much as we would get rid of most of the long AGPL notices.

Yajo:
> You should also include a copy of the license

should != must.

> multiple modules in a single repository could have different licenses. One single license file at repo root doesn't help with multiple licenses.

That can happen for sure, but also as a packager I'll tell you that it makes life much harder. The common approach is: you license your contributions under the same license the project is. If you plan to pull code with different licenses, you will endlessly enter other discussions about if licenses are compatible (because all GPL code infects their dependent code, as you probably know), about code headers (yet again), etc.

OCA has guidelines. IMHO one should be: *only AGPLv3+ code will be accepted*. Less problems for the future, believe me. :wink:

> multiple modules in a single repository could have different licenses.

Each file says its license anyway. No conflict here.

pedrobaeza:
About not including LICENSE file on each module: I agree with @Yajo. Link reference is enough.

About restricting license to AGPLv3, no. OCA accepts any OSI license compatible with the core. It's up to the contributor to choose the proper license to use. OCA will only check that it's OSI and that it's compatible. For example, website related module are more interesting to be LGPL instead of AGPL.

Yajo:
It was just a suggestion, but OCA's position is OK and very understandable. After all this, what about merging?

pedrobaeza:
There has been too many opinions, so we need the contrary ones to say if it's OK.

template/module/__init__.py
```diff
@@ -0,0 +1,3 @@
+# -*- coding: utf-8 -*-
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+# © <YEAR(s)> <AUTHOR(s)>
```
dreispt: I would prefer the copyright line to go first.

template/module/__openerp__.py
```diff
@@ -0,0 +1,26 @@
+# -*- coding: utf-8 -*-
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+# © <YEAR(s)> <AUTHOR(s)>
+
+# Instructions: https://www.odoo.com/documentation/8.0/reference/module.html
```
dreispt: Is this useful enough? I worry that people will leave it around then using the template.

template/module/__openerp__.py
```diff
@@ -0,0 +1,26 @@
...
+{
+    "name": "Module name",
+    "summary": "Module summary",
+    "version": "8.0.1.0.0",
+    "category": "Module category",
```
dreispt: I suggest using the `"Uncategorized"` value. Maybe add a comment at the end of the line with additional guidance.

template/module/__openerp__.py
```diff
@@ -0,0 +1,26 @@
...
+    "license": "AGPL-3",
+    "website": "https://odoo-community.org/",
+    "author": "Odoo Community Association (OCA)",
```
dreispt: Add a placeholder for the actual author name: 

    "author": "{author name}, Odoo Community Association (OCA)"

dreispt:
I believe the short form for the license header is OK, and honestly prefer it over the current situation.
Just made a couple of nitpick comments on the latest version of the code.

yvaucher:
I know should != must I was just pointing out the recommendations. I guess it could be kind of ok for version 8.0 as core modules are AGPL. However, I'm still concerned about the future for v 9.0 and I think it should be clear enough that we don't have LGPL for most if not all our modules. And I still think a link to a local file included in the module is better than an url (thinking of worst case scenarii).

By the way:
http://www.gnu.org/licenses/gpl-faq.html#WhyMustIInclude

This is more a must than a should.

This text might have been written a long time ago, but considering firewalls, outages, poor connectivity area and so on, it would be good to follow best practices.


For restriction of license OCA board stated this year that all compatible OSI license could be accepted. So it's a no for this kind of restriction.

pedrobaeza:
Well, as you said, this text was written a long time ago, and now Internet situation gets better. Anyways, the sole mention of the license name itself grants that you can check in any way the corresponding rights. It's like the laws: you're not including a copy of the law each time you mention it by its name. The name itself serves for recognizing that law, and can be search by that name. The same happens with AGPL-3.

dreispt:
Guys, IMO the inclusion or not of a LICENSE file is off topic for this PR.

Yajo:
https://github.com/OCA/maintainer-tools/pull/91#issuecomment-130323004 :+1:

I'll address the rest when I'm back from holidays. Happy summer!

rafaelbn:
IMO:

- These not needed as commented here: https://github.com/OCA/maintainer-tools/pull/91#discussion_r36850751

``` +# Instructions: https://www.odoo.com/documentation/8.0/reference/module.html ```

- And better like this as commented here: https://github.com/OCA/maintainer-tools/pull/91#discussion_r36850875

``` "author": "{author name}, Odoo Community Association (OCA)" ```

File: template/module/__openerp__.py (hunk 10 above)
Comment (Yajo) [inline template/module/__openerp__.py]: I agree.

**Comment by Yajo** [comment]
All done.

File: template/module/__init__.py (hunk 9 above)
Comment (dreispt) [inline template/module/__init__.py]: Can you place the Author and Order lines in the same order as in the `__openerp__.py`?

**Comment by dreispt** [comment]
A small nitpick but overall :+1:

**Comment by moylop260** [comment]
:+1:

**Comment by max3903** [comment]
:+1:

File: template/module/__init__.py (hunk 9 above)
Comment (moylop260) [inline template/module/__init__.py]: @dreispt 
Good point.
Maybe, should be a good idea use the same order in manifest key:
```python
    "website": "https://odoo-community.org/",
    "author": "<AUTHOR(S)>, Odoo Community Association (OCA)",
    "license": "AGPL-3",
    "application": False,
```
Just for satisfy our OCPD.

template/module/__openerp__.py
```diff
@@ -0,0 +1,24 @@
...
+# © <YEAR(S)> <AUTHOR(S)>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    "name": "Module name",
+    "summary": "Module summary",
+    "version": "8.0.1.0.0",
+    "category": "Uncategorized",
+    "license": "AGPL-3",
+    "website": "https://odoo-community.org/",
+    "author": "<AUTHOR(S)>, Odoo Community Association (OCA)",
```
moylop260: I a past discussion was remove the space between each author. (I don't have the reference now)

template/module/__openerp__.py
```diff
@@ -0,0 +1,24 @@
...
+    "application": False,
```
moylop260: Currently we don't have a rule to order keys...
But we have all modules with `installable` and `application` in last section.
I don't know if is a good idea change all modules with this new standard

template/module/__openerp__.py
Yajo: It's ugly without spaces, don't you think?

template/module/__openerp__.py
Yajo: My logic was to put first strings, then booleans, then lists. Seems cool as it is. Order does not matter in a dict, after all, and it's just a template, change it if you wish.

Yajo:
With https://github.com/grupoesoc/maintainer-tools/commit/ae2eb9904080f4d226a5b1f2e51535cabd2e6d8e I'm adding a dependency for `base` by default.

I know, I know, some will tell that it's not needed, but it is for one simple reason: updates.

When you update your code repositories in an Odoo server, updates are not applied by default. You must manually update each module, and it's not easy to know which ones were or not updated upstream. But there's one trick: update `base`, and all modules in your system will get updated. This happens because when you update one module, Odoo updates the ones that depend on it too. Modules not depending ultimately on `base` do not get updated that way, and that could lead to problems very easily.

BTW, this should be fixed in all OCA modules, and the trick should be documented somewhere... But that's another issue.

pedrobaeza:
@Yajo, that's not true: all modules depending directly or indirectly on base are updated if you update base, because there's a graph dependency resolution

yvaucher:
@Yajo no need for that explicit dependency even for --update=base

As an example, stock will be updated anyway if you installed it as it depends on product and product depends on base.

dreispt:
@pedrobaeza  @yvaucher You're right, still it's just a template and I think it's good to have it there.
Of course people are expected to replace it by more specific dependencies, if any.

yvaucher:
@dreispt agreed.

I was just arguing about the reason. :smile:  Of course no module should be orphan, so ok for having base by default.

moylop260:
My 2 comments was:

- Currently we don't have a rule to order keys of manifest
  But we have all modules with installable and application keys in last section.
  I don't know if is a good idea change all modules with this new standard

- I a past discussion was remove the space between each author. (I don't have the reference now)
  And our guidelines don't have space: https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#modules

dreispt:
> I don't know if is a good idea change all modules with this new standard

It's just a template, not a standard nor a mass change.

> In a past discussion was remove the space between each author

Initially no space is used; recently I've seen the space being added.
Does it make a difference? If it does it should be mentioned in the guidelines.

moylop260:
@dreispt 
IMHO this is a standard.
Because, you can use this template files to create a `oca-scaffold`
Similar to https://www.odoo.com/documentation/8.0/reference/cmdline.html#scaffolding

dreispt:
@moylop260 A template is a default. It's complies with standards but does not set standards nor excludes other compliant alternatives from being used.

Yajo:
@pedrobaeza @yvaucher Sorry, but my reason is legit. I double-checked today. Depending directly or indirectly on `base` does not matter. The problem is [depending on nothing, like `partner_relations`](https://github.com/OCA/partner-contact/blob/f99b1d18f69d48d5db5f94e18e414fea78a60a3a/partner_relations/__openerp__.py#L27). Install that module and then update `base`. You will see that it does not get updated. This way, the template implicitly says you: *if you depend on nothing, leave at least `base` there*.

dreispt:
Reviewers, are there any objections still to be addressed?

pedrobaeza:
@Yajo, you're right, you have to depend at least on base. That module is incorrect then and you should fill an issue. But that's one of the first things in the Odoo development guide. Anyway, that line on the template doesn't hurt, so we can keep it.

pedrobaeza:
For me :+1:

dreispt:
:+1: for me too

yvaucher:
Ok let's merge this
:+1: 

Thank you @Yajo

yvaucher:
So sad there is no copyleft symbol

PR #92, opened 2015-07-29
[REF] CONTRIBUTING.md: Add external dependencies case.

moylop260:
If we use `external_dependencies` we will have validation of installed packages in install module button.
How to work `external_dependencies` (without `static` folder): [here](https://asciinema.org/a/23098)

But if we don't use `try-except` we will have a error if exists `static` folder:
How to work `external_dependencies` (with `static` folder): [here](https://asciinema.org/a/23103)
Then we will need to add a try-except, but this should be a info error because the real error will show when you want to install the module if you don't use this module then you don't have warning or error just info log.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -258,6 +258,10 @@ except ImportError:
   _logger.info('Can not `import external_dependency_python_N`.')
 ```
 
+#### README
+If your module use extras dependencies of python or binaries, please explain how to install them in the `README.rst` file in the section `Installation`.
+
+
 ## Python
 
 ### PEP8 options
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -219,6 +219,45 @@ name.
 </record>
 ```
 
+### External dependencies
+
+#### `__openerp__.py`
+If your module use extras dependencies of python or binaries you should add to `__openerp__py` file the section `external_dependencies`.
+
+```python
+{
+    'name': 'Module of example',
+    'depends': ['base'],
+    ...
+    'external_dependencies': {
+        'bin': [
+            'external_dependency_binary_1',
+            'external_dependency_binary_2',
+            ...
+            'external_dependency_binary_N',
+        ],
+        'python': [
+            'external_dependency_python_1',
+            'external_dependency_python_2',
+            ...
+            'external_dependency_python_N',
+        ],
+    },
+    ...
+    'installable': True,
+}
+```
+
+#### ImportError
+In python files where you use a `import external_dependency_python_N` you will need to add a `try-except` with a info log.
+
+```python
+try:
+  import external_dependency_python_N
+except ImportError:
+  _logger.info('Can not `import external_dependency_python_N`.')
+```
+
 ## Python
 
 ### PEP8 options
@@ -603,6 +642,7 @@ The differences include:
     * Using one file per model
     * Separating data and demo data xml folders
     * Not changing xml_ids while inheriting
+    * Add guideline to use external dependencies
 * [Python](#python)
     * Fuller PEP8 compliance
     * Using relative import for local files
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -248,6 +248,7 @@ If your module use extras dependencies of python or binaries you should add to `
 ```
 
 An entry in `bin` needs to be in `PATH` identify with `which external_dependency_binary_N` command.
+
 An entry in `python` needs to be in `PYTHONPATH` identify with `python -c "import external_dependency_python_N"` command.
 
 #### ImportError
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -226,8 +226,7 @@ If your module use extras dependencies of python or binaries you should add to `
 
 ```python
 {
-    'name': 'Module of example',
-    'depends': ['base'],
+    'name': 'Example Module',
     ...
     'external_dependencies': {
         'bin': [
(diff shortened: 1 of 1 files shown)


max3903:
Can you also mention adding it into the module description (installation section) ?

https://github.com/OCA/maintainer-tools/tree/master/template/module#installation

Otherwise
:+1:

moylop260:
@max3903 
Thank for you comment.

Excuse me, I don't understand that you need that I add to https://github.com/OCA/maintainer-tools/tree/master/template/module#installation

Someone like as:
`Remeber add external_dependencies to your __openerp__.py`?

CONTRIBUTING.md
```diff
@@ -219,6 +219,45 @@ name.
 </record>
 ```
 
+### External dependencies
+
```
max3903: I mean add something like:

#### README.rst
If your module use extras dependencies of python or binaries, please explain how to install them in the `README.rst` file in the section `Installation`.

CONTRIBUTING.md
```diff
@@ -219,6 +219,45 @@ name.
...
+    ...
+    'installable': True,
+}
+```
+
+#### ImportError
+In python files where you use a `import external_dependency_python_N` you will need to add a `try-except` with a info log.
+
+```python
+try:
```
pedrobaeza: I prefer to enclose then into the method that uses this dependency, because it doesn't log anything if the module is uninstalled

dreispt: It's important to be at the import, because it's possible for the Odoo module can be imported even if not installed. The `try` will avoid server error in that case.

moylop260: @pedrobaeza 
You don't need add a validation in each method odoo not allow you to install it.
You can see next video: https://asciinema.org/a/23098

CONTRIBUTING.md
moylop260: Good point!

moylop260: Done!

CONTRIBUTING.md
moylop260: @pedrobaeza 
Maybe the merge of your comment is: 
Option 1: Change of `log.info` to `log.debug`
   You can choose to see this message.
Option 2: Change of `log.info` to `pass`
    You don't see log if is uninstalled, but if you try install it the you will see the error.

pedrobaeza: No, my option is to include the external dependency:

https://github.com/OCA/l10n-spain/blob/8.0/l10n_es_aeat/__openerp__.py#L48

and the import inside the method which needs it: 

https://github.com/OCA/l10n-spain/blob/8.0/l10n_es_aeat/wizard/export_to_boe.py#L58

This way, you don't receive any annoying log line with uninstalled modules, and when installing, module manager will check if you have the lib. You won't need either to make any check here, because when using method, you already assure that the library is available.

CONTRIBUTING.md
```diff
@@ -219,6 +219,49 @@ name.
...
+            'external_dependency_binary_2',
+            ...
+            'external_dependency_binary_N',
+        ],
+        'python': [
+            'external_dependency_python_1',
+            'external_dependency_python_2',
+            ...
+            'external_dependency_python_N',
+        ],
```
bwrsandman: Could you elaborate on what can go in bin and python, for example, an entry in bin needs to be in PATH? an entry in python is the name you import from or the name from pip freeze?

CONTRIBUTING.md
```diff
@@ -219,6 +219,49 @@ name.
...
 
+### External dependencies
+
+#### `__openerp__.py`
+If your module use extras dependencies of python or binaries you should add to `__openerp__py` file the section `external_dependencies`.
+
+```python
+{
+    'name': 'Module of example',
+    'depends': ['base'],
```
bwrsandman: Base is redundant.

CONTRIBUTING.md
```diff
@@ -219,6 +219,49 @@ name.
...
 ```
 
+### External dependencies
+
+#### `__openerp__.py`
+If your module use extras dependencies of python or binaries you should add to `__openerp__py` file the section `external_dependencies`.
+
+```python
+{
+    'name': 'Module of example',
```
bwrsandman: "Example Module"

CONTRIBUTING.md
moylop260: I like our standard of use import sorted in top file of python:
https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#imports

If you need use your library in many method you will need to add a `import`.

I undertand your point:
`you don't receive any annoying log line with uninstalled modules`
I have a proposal for you: https://github.com/OCA/maintainer-tools/pull/92/files#r35818714

pedrobaeza: I prefer what I say, but let others tell what they prefer.

lmignon: On my side I prefer to put `import` on top of the python file. The external dependency must always be included in the manifest.
My 2 cents

dreispt: @pedrobaeza not enclosing the import in a try block makes the dependency as good as required as long as the module is present in the addons path. So I strongly believe the try block is needed.
@lmignon Proposal in compatible with that you say.
@moylop260 IMO you should change the info to debug

moylop260:
Changes applied!

dreispt:
:+1:

CONTRIBUTING.md
yvaucher: I also prefer imports on top of file. As you might need your import in multiple methods.

yvaucher:
:+1:

PR #82, opened 2015-07-29
Variable name CamelCase potential problem of `redefined` variable.

moylop260:
I saw in oca guidelines this matter:
******************
#### Variable name :
*  **use camelcase for model variable**
* use underscore lowercase notation for common variable.
* since new API works with record or recordset instead of id list, don't suffix
  variable name with `_id` or `_ids` if they do not contain an id or a list of
  ids.

```python
ResPartner = self.env['res.partner']
partners = ResPartner.browse(ids)
partner_id = partners[0].id
```
******************

I want comment a problem with this:
 1. Use CamelCase in a variable don't is a pep8 standard.
 1. We will have potential problem of redefined variable.
   1. By example:
```python
        class ResPartner(model.Models):
        ...
        ResPartner = self.env['res.partner']
```

Why don't use as pep8 told us: snake_case to variable name?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -323,16 +323,25 @@ class account_invoice(orm.Model):
 ```
 
 #### Variable name :
-* use camelcase for model variable
-* use underscore lowercase notation for common variable.
+* use underscore lowercase notation for common variable (snake_case)
 * since new API works with record or recordset instead of id list, don't suffix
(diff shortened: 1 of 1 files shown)


moylop260:
PR generated.

bwrsandman:
:+1: Thank you!

pedrobaeza:
:+1:

PR #93, opened 2015-08-03
Include eggs folder on .gitignore, on MAc OS it is created with python setup.py install.

nhomar:


pedrobaeza:
I don't use Mac, but I suppose it can be. I prefer anyhow to add another line for explicitely say eggs instead of adding a wildcard that can exclude another files like egg_model.py

yvaucher:
@nhomar You know that you can setup a default gitignore file on you local config?

As @pedrobaeza it would be better to add an accurate name.

nhomar:
I close this. whatever! seriously.... (if in ANY os you generate a file with the setup.py the rule is include it in order to avoid miss commits)..

yvaucher:
@nhomar Your PR is valid it is just that it would be preferable to have:

    *.egg
    *.eggs

As *.egg* might be to wide.

nhomar:
@yvaucher done the change.

yvaucher:
@nhomar thanks
:+1:

pedrobaeza:
:+1:

dreispt:
:+1:

PR #94, opened 2015-08-03
Rename script to make s/README.md/README.rst

nhomar:


pedrobaeza:
I don't see this really needed. There is no README.md files currently in modules except the ones in the top directory for the repository description, which is intended, because that ones are with md syntax.

yvaucher:
Same opinion as @pedrobaeza 

We use `.rst` in modules because Odoo can only read this format. And me use `.md` on project root because it is the default format used by Github.

So :-1: on this.

(and sorry for the "in progress" tag I just deactivated that on waffle.io)

nhomar:
When you have branches with hundreds of modules like ours, this can help to standarize the repository previously started to migrate OCA related.

To avoide comments like this: https://github.com/OCA/rma/pull/26#issuecomment-127173669 which do not add any value but are necesary to take into account because we need to comply with standards.

IMHO, if you need to do something twice automate it that's a sysadmin rule since always.

Now we have yet thousands of people that if we help them with tools to include them here it helps.

But BTW, I close the issue. no problem, (if you need it it will be in vauxoo's copy of this module).

PR #95, opened 2015-08-03
Replace Module Name by a placeholder so less people will forget it, h…

yvaucher:
…opefully

I guess that as all section title are already filled people thinks "Module name" is already a section or forget to replace it afterwards. Let's try with something closer to a placeholder to replace.


pedrobaeza:
:+1: let's hope this will be enough

dreispt:
:+1:

PR #98, opened 2015-08-04
[IMP] Add demo section

max3903:


moylop260:
IMHO This documentation should has in root README of project.

max3903:
The README at the root of the repo is not displayed on apps.odoo.com.

Having it in each module will allow anyone to easily test it in just a click away from apps.

moylop260:
@max3903 
Thanks for you explaination.
I'm agree with you!

nhomar:
+1

dreispt:
The idea is nice, but I dislike the additional README section.
I suggest a button / link at the top of the README with something like "Try it live".

dreispt:
I don't think the runbot link is working. For example:https://runbot.odoo-community.org/runbot/project-service/8.0

nhomar:
2015-08-05 2:32 GMT-05:00 Daniel Reis <notifications@github.com>:

> I don't think the runbot link is working. For example:
> https://runbot.odoo-community.org/runbot/project-service/8.0
>

I like the link (alà travis) and it is necessary to navigate quickly (we
try to use it in our modules).

But, a good little image beside travis links should be enought not all a
full section.

Regards.


-- 
--------------------
Saludos Cordiales

CEO at Vauxoo <https://www.vauxoo.com> Odoo's Gold Partner.

[image: --]
Nhomar Hernandez
[image: http://]about.me/nhomar
<http://about.me/nhomar?promo=email_sig>

pedrobaeza:
Runbot use the id in the URL to access to the branch. For example, https://runbot.odoo-community.org/runbot/repo/github-com-oca-l10n-spain-189, so there's no way to predict what is this URL. We can have an index, but it's tedious to maintain

nhomar:
2015-08-05 2:55 GMT-05:00 Pedro M. Baeza <notifications@github.com>:

> Runbot use the id in the URL to access to the branch. For example,
> https://runbot.odoo-community.org/runbot/repo/github-com-oca-l10n-spain-189,
> so there's no way to predict what is this URL. We can have an index, but
> it's tedious to maintain
>

Yes you are all right, I will trigger the feature of runbot to have a
search api on our side.

One time we have the feature we come back here (just in this moment we can
discuss aroud that).


-- 
--------------------
Saludos Cordiales

CEO at Vauxoo <https://www.vauxoo.com> Odoo's Gold Partner.

[image: --]
Nhomar Hernandez
[image: http://]about.me/nhomar
<http://about.me/nhomar?promo=email_sig>

max3903:
The URL uses the repo id in Runbot, which we know when we add it here:
https://github.com/OCA/maintainer-tools/blob/master/tools/repos_with_ids.txt

dreispt:
@max3903 OK, but I think what only a few projects have running instances available.

max3903:
So it will motivate us to make them work.

By the way, this is the template, not a mass commit on every module to add the link. The link can be added to modules which have a working runbot progressively.

pedrobaeza:
Yeah, I agree, but then the template should include a link to that index, and build correctly the link (it's not only the id). I'm afraid we will need as reviewers to check this and correct it in most cases.

max3903:
@pedrobaeza pretty much the same as the link to Github to report issues

pedrobaeza:
Wow, I was thinking about this kind of link: https://runbot.odoo-community.org/runbot/repo/github-com-oca-account-analytic-87, but I see that the format you have put allow to access directly to the instance, and even you don't need to put admin/admin, so :+1: to this method, but please remove the need of putting credentials, and link to the repo ids list.

max3903:
@pedrobaeza @dreispt How about this?

pedrobaeza:
Perfect! Maybe @nhomar's solution can be another one (write a controller that use the name instead of the id), but for now, I see it correct.

moylop260:
We solved it with a index:
http://runbot.vauxoo.com/runbot/team/ovl-29#98pull_4
Notice: #98pull_4
But our runbot has a multi-branch configuration by environment (not by repository) and have hardcode changes... [large history](https://github.com/odoo/odoo-extra/pull/29#issuecomment-51511242)
Maybe with native runbot you can add a index by branch:

https://runbot.odoo-community.org/runbot/repo/github-com-oca-account-analytic-87#8.0
Notice: #8.0

Or you can use the `search` url:
https://runbot.odoo-community.org/runbot/repo/github-com-oca-account-analytic-87?search=8.0

dreispt:
To be honest I think the link will be lost in the clutter.
I would rather have a button at the top, or maybe a ribbon.

Something similar to these cloud deploy Examples:

- https://blog.openshift.com/updated-openshift-deploy-buttons
- https://github.com/seven1m/do-install-button/blob/master/README.md
- https://github.com/heroku/button-sample/blob/master/README.md

Until someone skillfull designs a nice button, a simple button would do the trick.

max3903:
@dreispt Button image is cooking. Should be ready by tomorrow.

max3903:
Button image is ready, courtesy of http://www.pixelove.ca

pedrobaeza:
Cool, the button looks awesome!

Just one thing: I think a better place is just after the "Usage" section.

max3903:
@pedrobaeza If the button is before the instructions, I am afraid people will not read them. Read first on how to do it and then try to do it.

pedrobaeza:
OK, you're right :+1:

lmignon:
:+1: Great idea an nice button! Thank you @max3903

dreispt:
@max3903 Do you really believe that people will read the instruction *before* hitting the button just because it's placed there? :wink: 
IMO it deserves to be at the top, just before the `Installation` section.
People can have a first impression navigating it, and then have a closer look at the usage.

max3903:
@dreispt I prefer to encourage people to read first. If you want to put the logo up in your module, so be it. This is just a template.

dreispt:
Fine. Great button BTW!
:+1:

moylop260:
I'm late.

But the link
`https://runbot.odoo-community.org/runbot/{repo_id}/{branch}`
Told me that generate a next example:
https://runbot.odoo-community.org/runbot/github-com-oca-account-analytic-87/8.0

The correct template link should be:
`https://runbot.odoo-community.org/runbot/repo/{repo_url_id}?search={branch_name}`

To generate next link
https://runbot.odoo-community.org/runbot/repo/github-com-oca-account-analytic-87?search=8.0

pedrobaeza:
@moylop260, instructions are correct. Try for example https://runbot.odoo-community.org/runbot/92/8.0

PR #101, opened 2015-08-06
[REF] CONTRIBUTING: Add use of default label string in fields.

moylop260:
By default new api use name of field in label string.
Use this feature if is posible.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -400,6 +400,9 @@ class...
   (example: sale_order_line_ids)
 * `Many2One` fields should have `_id` as suffix
   (example: partner_id, user_id, ...)
+* If the technical name of the field (the variable name) is the same to the string of the label, don't put `string` parameter for new API fields, because it's automatically taken. If your variable name contains "_" in the name, they are converted to spaces when creating the automatic string and each word is capitalized.
+  (example: old api `'name': fields.char('Name', ...)`
+            new api `'name': fields.Char(...)`)
 * Method conventions
     * Compute Field: the compute method pattern is `_compute_<field_name>`
     * Search method: the search method pattern is `_search_<field_name>`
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -400,6 +400,10 @@ class...
   (example: sale_order_line_ids)
 * `Many2One` fields should have `_id` as suffix
   (example: partner_id, user_id, ...)
+* If the name of field is the same to name of label string don't setting `string` parameter for fields of api v8.
```
pedrobaeza: Rephrase this to:

`If the technical name of the field (the variable name) is the same to the string of the label, don't put `string` parameter for new API fields, because it's automatically taken. If your variable name contains "_" in the name, they are converted to spaces when creating the automatic string and each word is capitalized`

moylop260: @pedrobaeza 
Thank for you feedback.
Done!

CONTRIBUTING.md
```diff
@@ -400,6 +400,9 @@ class...
   (example: sale_order_line_ids)
 * `Many2One` fields should have `_id` as suffix
   (example: partner_id, user_id, ...)
+* If the technical name of the field (the variable name) is the same to the string of the label, don't put `string` parameter for new API fields, because it's automatically taken. If your variable name contains "_" in the name, they are converted to spaces when creating the automatic string and each word is capitalized.
```
pedrobaeza: You have to enclose the _ like code, because it's making the parser to interpret it as italics begin

moylop260: I don't understand
Can you help to clarify?
FYI I saw this: http://awesomescreenshot.com/0d954qsy30

pedrobaeza: Forget it, it was, only something of the preview. That doesn't happen in the full file view.

pedrobaeza:
:+1:

nhomar:
:+1:

dreispt:
:+1:

lmignon:
:+1:

PR #105, opened 2015-08-13
[ADD] oca-autopep8: Add script autopep8 extended to support replace off class name from snake_case to CamelCase.

moylop260:


moylop260:
Somebody here?

README.md
```diff
@@ -76,6 +76,23 @@ This tools will also warn you what are the specific labels on some repository
...
 
+
+**Auto fix pep8 guidelines**
+
+To auto fix pep8 guidelines of your code you can run:
+
+    $ oca-autopep8 -ri PATH
+
+This script overwrite with monkey patch the original script of [autopep8](https://github.com/hhatto/autopep8)
+to support custom refactory of code.
```
pedrobaeza: s/refactory of code/code refactoring

README.md
```diff
@@ -76,6 +76,23 @@ This tools will also warn you what are the specific labels on some repository
...
+
+* List of errors added:
+
+    - `CW0001` Named of class has snake_case style, should use CamelCase.
```
pedrobaeza: s/Named of class/Class name

moylop260:
@pedrobaeza 
Thanks you for comment
Now, Fix typo

pedrobaeza:
Thanks :+1:

tools/autopep8_extended.py
```diff
@@ -0,0 +1,192 @@
...
+                    old_class_name=class_original_name,
+                    new_class_name=class_renamed[
+                        class_original_name]['renamed'],
+                    lines_columns=class_renamed[
+                        class_original_name]['line_col']
+                ),
+            })
+        return check_result
+
+    def _execute_pep8_extendend(self):
```
nhomar: Document please.

tools/autopep8_extended.py
```diff
@@ -0,0 +1,192 @@
...
+    def get_checks(self):
+        msgs = {
+            'CW0001': 'Class name with snake_case style found, '
+                      'should use CamelCase. Change of "{old_class_name}" '
+                      'to "{new_class_name}" next (lines,columns): '
+                      '{lines_columns}',
+        }
+        return msgs
+
+    def check_cw0001(self):
```
nhomar: Document please

nhomar:
Just adding the documented stuff.

:+1:

tools/autopep8_extended.py
```diff
@@ -0,0 +1,192 @@
...
+                node_renamed = inflection.camelize(
+                    node.name, uppercase_first_letter=True)
+                renamed_names.append(node_renamed)
+                if node_renamed != node.name:
+                    class_renamed.setdefault(
+                        node.name, {'line_col': []})
+                    class_renamed[node.name]['renamed'] = \
+                        node_renamed
+                    class_renamed[node.name]['line_col'].append((
+                        node.lineno + line_deleted,
```
lmignon: Even if in python 1 + True = 2, IMO, the line_deleted should not be a boolean.

tools/autopep8_extended.py
```diff
@@ -0,0 +1,192 @@
...
+        }
+        return msgs
+
+    def check_cw0001(self):
+        code = 'CW0001'
+        msg = self.get_checks()[code]
+        class_renamed = {}
+        check_result = []
+
+        # Fix 'SyntaxError: encoding declaration in Unicode string'
```
lmignon: IMO you should split your method into smaller methods...
For exemple, this part of code removing the encoding comment found in first two lines should be externalized in an other method:  
```python
def strip_encoding():```
    """Remove encoding comment if found in first two lines
        If the first or second line has the `# xx coding xx` comment,
        it will be removed.
    """
```

tools/autopep8_extended.py
```diff
@@ -0,0 +1,192 @@
...
+        parsed_source = self.source
+        line_deleted = False
+        if len(parsed_source) >= 1 and \
+           ('# -*- coding:' in parsed_source[0] or
```
lmignon: Not sure if you cover all the ways you can declare the encoding.. Moreover the detection logic is duplicated...
What do you think of
```python
lines = parsed_source[0:2]
for index, line in enumerate(lines):
    if line.startswith('#') and 'conding:' in line:
        line_deleted = 1
        parsed_source = parsed_source[0:index] + parsed_source[index+1:]
```

lmignon:
:-1:  IMO We should avoid to monkey patch autopep8. You should do a PR on https://github.com/hhatto/autopep8.

pedrobaeza:
@lmignon I don't think we get accepted on autopep8 this specific odoo rules, but he can try...

lmignon:
@pedrobaeza We should at least make a try and if it's rejected we can reconsider adding it to the oca toolbox or even better create a new dedicated python lib. This kind of need is not unique to OCA and IMO we should avoid to build/hack tools making OCA a kind of new divergent python community.

pedrobaeza:
OK, let's try

moylop260:
Thanks for your comments.

FYI why monkey patch: https://github.com/hhatto/autopep8/issues/214

lmignon:
@moylop260 Thank you for your attempt to contribute to autopep8. Nevertheless, it seems that they are not ready to integrate this improvement in the official distribution. https://github.com/hhatto/autopep8/issues/214#issuecomment-135226535
Merging this change in MT has the advantage to provide a quick response to a recurring need. But IMO it would be better to create a new project under OCA `???autopep8???` to be able to distribute this enhancement in a dedicated python lib. 
Whichever way of distributing this library what do you think of my 2 comments (https://github.com/OCA/maintainer-tools/pull/105#discussion_r37952876, https://github.com/OCA/maintainer-tools/pull/105#discussion_r37953610)

moylop260:
@lmignon 

IMHO 

> Merging this change in MT has the advantage to provide a quick response to a recurring need. But IMO it would be better to create a new project under OCA `???autopep8???` to be able to distribute this enhancement in a dedicated python lib. 

My vote to use this project for `oca-autopep8`

> Whichever way of distributing this library what do you think of my 2 comments (https://github.com/OCA/maintainer-tools/pull/105#discussion_r37952876, https://github.com/OCA/maintainer-tools/pull/105#discussion_r37953610)

You feedback is very good!
Thanks for that.
I will work it

moylop260:
@lmignon & @nhomar 
 - [x] Methods documented
 - [x] Added new method to strip unicode
 - [x] Added unitests

nhomar:
After all the discussions i give my :+1: 
The only point stay in the grey area is if we use a separate python package or not, I think for now it can be here and when we decide do such job then we can separate it but in a future.

.travis.yml
```diff
@@ -6,13 +6,14 @@ python:
     - "2.7"
 
 install:
-    - pip install coveralls flake8
+    - pip install coveralls flake8 unittest2
```
lmignon: Why do you need `unittest2`? Maybe I'm wrong, but the result will be the same if your test class extends `unittest.TestCase` in place of `unittest2.TestCase`

.travis.yml
```diff
@@ -6,13 +6,14 @@ python:
...
     - pip install -r requirements.txt
 
 script:
     - python setup.py flake8
     - coverage run setup.py test
     - python setup.py install
+    - coverage run tests/test_autopep8_extended.py
```
lmignon: This line is not required. You should modify the setup.py to looks like:
```
import setuptools

setuptools.setup(setup_requires=['pbr'],
                 pbr=True,
                 test_suite="tests",
                 package_data={'': ['*.yaml']})
```

lmignon:
@moylop260 Thank you for all your work. IMO 2 issues remains. Your should remove the dependency on unittest2 and avoid to explicitly call your tests in travis.yml by adding the `test_suite="tests"` in setup.py. I've tested these modifications locally and your tests are executed when calling `coverage run setup.py test`
```
....
test_camelcase (tests.test_autopep8_extended.TestAutopep8Extended)
Test normal case of snake_case to CamelCase ... ok
test_coding (tests.test_autopep8_extended.TestAutopep8Extended)
Test coding and encoding lines. ... ok
test_wrong_camelcase (tests.test_autopep8_extended.TestAutopep8Extended)
Test wrong case when exists same class name two times ... Waning: Aborted. You have two class named my_name and MyName in a same file.
Review your next class names: ['HelloWorld', 'HelloWorld']
ok

----------------------------------------------------------------------
Ran 3 tests in 0.044s

moylop260:
@lmignon 
Good feedback!
Thank you!

Fixed!

lmignon:
@moylop260 Once again thank you. :+1:

moylop260:
3 approvals and PR of 18 days ago
Then merging...

PR #111, opened 2015-08-28
[REF] add link to AGPL licence;

legalsylvain:


max3903:
:+1:

yvaucher:
:+1:

pedrobaeza:
:+1:

pedrobaeza:
@lmignon has mentioned on a PR that this badge is not shown in module form view and he proposes to move below the title

lmignon:
@pedrobaeza You're right :smirk: Take a look to the app store and you'll see that the licence badge is missing on all OCA modules (except for miss_builder)

yvaucher:
discussed here: https://github.com/OCA/sale-workflow/pull/54#discussion_r38218235

PR #117, opened 2015-09-06
Add the inter-company repo for inter-company processes via connector.

elicoidal:
Changes to be committed:
	modified:   oca_projects.py
	modified:   repos_with_ids.txt


pedrobaeza:
:+1:

sbidoul:
:+1:

guewen:
:+1: for the PR
:-1: for the name of the project (I didn't find where I could comment about that): a  connector addon for synchronizing odoo to odoo is not well advertised with `inter-company` IMO. We have projects that use Odoo as a framework, absolutely not using the "company" thing. In that context, we are not doing company->company but odoo->odoo and if we were looking for a connector odoo to odoo in such context we would not look for an `inter-company` project.

elicoidal:
Actually there has been some discussions about it and functionally the original idea was to be able to manage inter-company transactions (hence the name).
Using the connector is a tool to achieve this but I am not sure whether it is relevant to add "connector" or switch to "Odoo2Odoo connector"

guewen:
The original idea does not preclude other uses cases ;-) Again, if someone searches a tool for Odoo->Odoo synchronisations but hasn't the 'inter-company' idea in head, he won't search in this project. Anyway I won't argue partly because I didn't see if the module was generic or had dependencies on SO/PO and partly because we have to put a name on the repo and there is no perfect name.

elicoidal:
True !
Actually those 2 ideas are mixed inside each other... and might deserve 
to separate them into 2 different repos:
(1) odoo2odoo: master data synchronization (generic connector)
(2) Intercompany: using (1) for the master data and more specific for 
SO2PO, DO2IS etc.

dreispt:
> for the name of the project (I didn't find where I could comment about that)

Following a similar discussion, IMO repo creation should be decided by the board *after* being discussed in the contributors ML.

guewen:
> Actually those 2 ideas are mixed inside each other... and might deserve to separate them into 2 different repos: (1) odoo2odoo: master data synchronization (generic connector) (2) Intercompany: using (1) for the master data and more specific for SO2PO, DO2IS etc.

:+1:

elicoidal:
Actually it has been discussed in the contributor mailing list (check 
Odoo 2 Odoo subject), mostly about separation between multi-company and 
inter-company, involving Joel and myself.

pedrobaeza:
IMO this one should be named with connector-odoo2odoo nomenclature and put here the base synchronization and let in multi-company repo the modules for the inter-company specific thing, with the dependency to this one.

elicoidal:
>IMO this one should be named with connector-odoo2odoo nomenclature and put here the base >synchronization and let in multi-company repo the modules for the inter-company specific thing, with >the dependency to this one.
That could be a good idea (renaming this repo into connector-odoo2odoo).
I would suggest to bring this discussion back to the contributors list and copy Malte the PR

maljac:
> Actually those 2 ideas are mixed inside each other... and might deserve
> to separate them into 2 different repos:
> (1) odoo2odoo: master data synchronization (generic connector)
>(2) Intercompany: using (1) for the master data and more specific for
> SO2PO, DO2IS etc.

From my point of view, totaly correct. But, as I already wrote in the mail list, in my oppinion we should not mix intercompany process addons with multi-company process addons. 

When I was new to Odoo, I felt in that "trap" and thought Odoo is able to manage intercompany (in terms of multiple odoo instances). This is because of the Odoo terminology "multi-company", which is actually should be multitenant.

So, I think the most correct separation of repos would be:

- ``multi-company``: As it is, managing process within same odoo instance
- ``connector-odoo2odoo``: As the name suggests, basic features and ability to provide o2o communication (as @guewen suggested)
- ``intercompany``: Process specific addons for ic processes, depending on connector-odoo2odoo (as @elicoidal suggested)

pedrobaeza:
Join both types of modules (multi- and inter-company) is only for administrative convenience, to avoid an overload managing repositories, because both subjects are very limited in scope. We can make that separation if all the board agree to have another repo, or rename multi-company to multi-inter-company to have clear the concepts involved but don't create another one repo. What do you think?

maljac:
I agree about the administrative effort. I am quite new to the OCA process, but it makes sense to keep numbers of repos low. 

If thats the point, I would suggest creating the ``connector-odoo2odoo`` repository. This repository will contain the base connector as well as process related addons. Since the process specific addons will be highly dependent on the ``base_odoo2odoo`` this makes sense, similar to the structure in  https://github.com/OCA/connector-telephony.

But if all you have different opinions about that, I am also fine to split into ``connector-odoo2odoo`` and ``inter-multi-company``.

elicoidal:
May I suggest to keep current multi-company as is and modify the readme file to explain the object and change the name of the current inter-company to connector-odoo2odoo.
We should have write a blog post as well to communicate on the role of each of them

maljac:
@elicoidal :+1:

elicoidal:
I have modified the github organization according to new naming.
I will create a new commit

PR #118, opened 2015-09-06
update oca_projects.py

sbidoul:


pedrobaeza:
:+1:

As this is a typo, I fast-track it.

PR #121, opened 2015-09-07
Fixed contributing import rules conflict

ajite:
There was a conflict between these two rules when importing a third party lib due to the `try-except` with the `logger`:

**Rule 1:**
> The imports are ordered as

> 1. Externals libs (One per line sorted and splitted in python stdlib)
> 2. Imports of `openerp`
> 3. Imports from Odoo modules (rarely, and only if necessary)
> 4. Local imports in the relative form

> Inside these 4 groups, the imported lines are alphabetically sorted.

**Rule 2:** 

> In python files where you use a `import external_dependency_python_N` you will need to add a `try-except` with a debug log.

> ```python
 try:
  import external_dependency_python_N
except ImportError:
  _logger.debug('Can not `import external_dependency_python_N`.')
```

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -322,27 +322,35 @@ a few exceptions:
 
 The imports are ordered as
 
-1. Externals libs (One per line sorted and splitted in python stdlib)
-2. Imports of `openerp`
-3. Imports from Odoo modules (rarely, and only if necessary)
-4. Local imports in the relative form
+1. Standard library imports
+2. Related third party imports (One per line sorted and splitted in python stdlib)
+3. Odoo imports (`openerp`)
+4. Imports from Odoo modules (rarely, and only if necessary)
+5. Local imports in the relative form
 
-Inside these 4 groups, the imported lines are alphabetically sorted.
+Inside these 5 groups, the imported lines are alphabetically sorted.
 
 ```python
 # 1: imports of python lib
 import base64
+import logging
+_logger = logging.getLogger(__name__)
 import re
 import time
-# 2:  imports of openerp
+# 2:  import of third party lib
+try:
+  import external_dependency_python_N
+except ImportError:
+  _logger.debug('Can not `import external_dependency_python_N`.')
+# 3:  imports of openerp
 import openerp
 from openerp import api, fields, models  # alphabetically ordered
 from openerp.tools.safe_eval import safe_eval as eval
 from openerp.tools.translate import _
-# 3:  imports from odoo modules
+# 4:  imports from odoo modules
 from openerp.addons.website.models.website import slug
 from openerp.addons.web.controllers.main import login_redirect
-# 4: local imports
+# 5: local imports
 from . import utils
 ```
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -323,10 +323,11 @@ a few exceptions:
 The imports are ordered as
 
 1. Standard library imports
-2. Related third party imports (One per line sorted and splitted in python stdlib)
+2. Known third party imports (One per line sorted and splitted in python stdlib)
 3. Odoo imports (`openerp`)
 4. Imports from Odoo modules (rarely, and only if necessary)
 5. Local imports in the relative form
+6. Unknown third party imports (One per line sorted and splitted in python stdlib)
 
 Inside these 5 groups, the imported lines are alphabetically sorted.
 
@@ -334,24 +335,31 @@ Inside these 5 groups, the imported lines are alphabetically sorted.
 # 1: imports of python lib
 import base64
 import logging
-_logger = logging.getLogger(__name__)
 import re
 import time
-# 2:  import of third party lib
-try:
-  import external_dependency_python_N
-except ImportError:
-  _logger.debug('Can not `import external_dependency_python_N`.')
+
+# 2: import of known third party lib
+import lxml
+
 # 3:  imports of openerp
 import openerp
 from openerp import api, fields, models  # alphabetically ordered
-from openerp.tools.safe_eval import safe_eval as eval
+from openerp.tools.safe_eval import safe_eval
 from openerp.tools.translate import _
+
 # 4:  imports from odoo modules
 from openerp.addons.website.models.website import slug
 from openerp.addons.web.controllers.main import login_redirect
+
 # 5: local imports
 from . import utils
+
+# 6: Import of unknown third party lib
+_logger = logging.getLogger(__name__)
+try:
+  import external_dependency_python_N  ########### 3rd party here?
+except ImportError:
+  _logger.debug('Can not `import external_dependency_python_N`.')
 ```
 
  * Note:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -329,7 +329,7 @@ The imports are ordered as
 5. Local imports in the relative form
 6. Unknown third party imports (One per line sorted and splitted in python stdlib)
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by ajite** [comment]
@oca-clabot (or to whoever handles the bot): I signed last year.

See issue #119 for that branch

**Comment by moylop260** [comment]
What about `isort` command?
How to order with this case?

**Comment by guewen** [comment]
@moylop260 
Doesn't `isort` already sort the imports the same? This order comes from `pep8`.

:+1:

**Comment by dreispt** [comment]
:+1:

**Comment by moylop260** [comment]
I executed isort in next code:
```python
# 1: imports of python lib
import base64
import logging
_logger = logging.getLogger(__name__)
import re
import time
# 2:  import of third party lib
try:
  import external_dependency_python_N
except ImportError:
  _logger.debug('Can not `import external_dependency_python_N`.')
# 3:  imports of openerp
import openerp
from openerp import api, fields, models  # alphabetically ordered
from openerp.tools.safe_eval import safe_eval as eval
from openerp.tools.translate import _
# 4:  imports from odoo modules
from openerp.addons.website.models.website import slug
from openerp.addons.web.controllers.main import login_redirect
# 5: local imports
from . import utils
```

And the result was:
```python
# 1: imports of python lib
import base64
import logging
import re
import time

import openerp
# 3:  imports of openerp
from openerp import api, fields, models  # alphabetically ordered
from openerp.addons.web.controllers.main import login_redirect
# 4:  imports from odoo modules
from openerp.addons.website.models.website import slug
from openerp.tools.safe_eval import safe_eval as eval
from openerp.tools.translate import _

# 5: local imports
from . import utils

_logger = logging.getLogger(__name__)
# 2:  import of third party lib
try:
  import external_dependency_python_N
except ImportError:
  _logger.debug('Can not `import external_dependency_python_N`.')
```

```diff
diff --git imports.py imports.py.original
index 617909a..94f13ff 100644
--- imports.py
+++ imports.py.original
@@ -1,24 +1,21 @@
 # 1: imports of python lib
 import base64
 import logging
+_logger = logging.getLogger(__name__)
 import re
 import time
-
-import openerp
+# 2:  import of third party lib
+try:
+  import external_dependency_python_N
+except ImportError:
+  _logger.debug('Can not `import external_dependency_python_N`.')
 # 3:  imports of openerp
+import openerp
 from openerp import api, fields, models  # alphabetically ordered
-from openerp.addons.web.controllers.main import login_redirect
-# 4:  imports from odoo modules
-from openerp.addons.website.models.website import slug
 from openerp.tools.safe_eval import safe_eval as eval
 from openerp.tools.translate import _
-
+# 4:  imports from odoo modules
+from openerp.addons.website.models.website import slug
+from openerp.addons.web.controllers.main import login_redirect
 # 5: local imports
 from . import utils
-
-_logger = logging.getLogger(__name__)
-# 2:  import of third party lib
-try:
-  import external_dependency_python_N
-except ImportError:
-  _logger.debug('Can not `import external_dependency_python_N`.')
```

What is wrong here?

moylop260:
Maybe we need to split:
 - # 2:  know import of third party lib (into of oficial requirements.txt of odoo/oca)
 - # 6:  unknow import of third party lib (required a external-lib in manifest and ImportError)

ajite:
@moylop260 : I am not sure because it would not respect the PEP8 convention. Local imports should be imported after any third-party libraries.

moylop260:
Yes, I'm agree with you.
And all import's in top section.
`_logger` should be after import's section.

But we have a special case that don't is defined in pep8 rules, by example:
 - All `import` in top section.
 - `_logger` should be `LOGGER` because is a global variable.
 - We should avoid a `try-ImportError` sentence.

But we have next issue https://github.com/odoo/odoo/issues/7507 that generate other problems.
Then we need to define how to solve it

My vote to use the proposal of isort.

ajite:
@moylop260: isort would indeed import these "unknown" third party libraries at the end. I am fine with both ways. Let's wait for some more feedbacks.

dreispt:
@moylop260 I guess `isort` can't "properly" handle imports in a try block, and treats them as regular code.

moylop260:
@dreispt 
Yes, I'm agree with you.
Same case with pep8.

`Imports are always put at the top of the file, just after any module comments and docstrings, and before module globals and constants.`
But we need a global variable: `logger`
But we need all `import` in top of the file.

Then we have a problem with this special case.

dreispt:
Let the logger global be a special case then.
And I agree that it should be in uppercase: `_logger` -> `LOGGER`

guewen:
> And I agree that it should be in uppercase: _logger -> LOGGER

I don't think so. Constants should be in capital letters, but _logger isn't really a constant. It seems to be a peculiarity of pylint to consider all module-wide variables to be constants and we find several workarounds for this issue (not necessarily good) when searching for that on the web (examples https://groups.google.com/forum/#!topic/comp.lang.python/u8Lf2saMStQ, http://stackoverflow.com/questions/10815549/pylint-showing-invalid-variable-name-in-output) .

dreispt:
@guewen You're right: uppercase names are for constants, not global variables.

ajite:
Hi all,

Thank you for your feedbacks.

So what is the status on that? Do we all agree or do I need to change something :)?

moylop260:
We have a recursive conflict.

Option | Fix | Conflict
----- | ------- | -------------
A | Top import | 3rd party lib
B | 3rd party lib | Top import

Option A:
```python
# 1: imports of python lib
import base64
import logging
import re
import time

# 2:  know import of third party lib
import lxml

# 3:  imports of openerp
import openerp
from openerp import api, fields, models  # alphabetically ordered
# 4:  imports from odoo modules
from openerp.addons.web.controllers.main import login_redirect
from openerp.addons.website.models.website import slug
from openerp.tools.safe_eval import safe_eval as eval
from openerp.tools.translate import _

# 5: local imports
from . import utils

# 6:  Unknow import of third party lib
_logger = logging.getLogger(__name__)
try:
  import external_dependency_python_N  ########### 3rd party here?
except ImportError:
  _logger.debug('Can not `import external_dependency_python_N`.')
```

Option B
```python
# 1: imports of python lib
import base64
import logging
_logger = logging.getLogger(__name__)  ########### Variable here ?
import re
import time
# 2:  import of third party lib
try:
  import external_dependency_python_N
except ImportError:
  _logger.debug('Can not `import external_dependency_python_N`.')
# 3:  imports of openerp
import openerp
from openerp import api, fields, models  # alphabetically ordered
from openerp.tools.safe_eval import safe_eval as eval
from openerp.tools.translate import _
# 4:  imports from odoo modules
from openerp.addons.website.models.website import slug
from openerp.addons.web.controllers.main import login_redirect
# 5: local imports
from . import utils
```

What you choose?
@all vote is welcome !

+1 with Option A

guewen:
Agree with A.

Just a side-note remark about your example, on this line:

```python
from openerp.tools.safe_eval import safe_eval as eval
```
It is better not to alias `safe_eval` and keep
```python
from openerp.tools.safe_eval import safe_eval
```
so when reading code we aren't forced to go up to the imports check if the `eval` is well a `safe_eval`

ajite:
:+1: with A. I will change my PR :)

dreispt:
No strong opinion, but :+1: for A

moylop260:
+1 with @guewen `safe_eval` comment (My bad I used current copy&paste)

moylop260:
Changed applied

:+1:

guewen:
> (My bad I used current copy&paste)

It was indeed not yours but already in the original example.

PR #123, opened 2015-09-12
[FIX] Work around Transifex/Rackspace limitations

max3903:


max3903:
Seems to be working:
https://github.com/oca-transbot

moylop260:
@max3903 
Can you explain a little more this patch?

max3903:
@moylop260 Sure. We hit some Rackspace limitations:
http://docs.rackspace.com/loadbalancers/api/v1.0/clb-devguide/content/Rate_Limits-d1e821.html

so I paused the script at the end of a file and at the end of a project.

moylop260:
Can you use cron?
https://github.com/OCA/maintainer-tools/blob/master/Dockerfile-tx-bot#L38-L40

max3903:
@moylop260 I already do. The script is run once a week. The problem is that it connects too often to Transifex during the execution.

pedrobaeza:
Please add a little comment over each sleep to tell about the problem, but :+1:

pedrobaeza:
For references, see this section in Transifex API doc: http://docs.transifex.com/api/#api-rate-limits. The one that is struggling us is the throttling in Django.

lmignon:
:+1:

sbidoul:
:+1: merging since it's running fine in production.

PR #124, opened 2015-09-14
[IMP] complete coding message

bealdav:
As far I know, OCA decided to stick to python standard (a real good thing).
In coding matter, we should allowed to add this string too.
In the official doc https://www.python.org/dev/peps/pep-0263 § Defining the Encoding, 
it's the first example (and the shorter) that we could use.

@moylop260 asked me to make this PR 
https://github.com/OCA/server-tools/pull/171#discussion_r38918913

I know, somebody will tell us: all oca modules use -\*-  ...  -\*-
But last year, classes in OCA were not in camel case and we changed 
to stick standards and get it easier to read.

And to conclude it's good thing for the earth, you save bytes in the cloud :earth_africa: ;-)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -368,7 +368,7 @@ except ImportError:
 
 ### Idioms
 
-* Each python file should have ``# -*- coding: utf-8 -*-`` as first line
+* Each python file should have ``# coding: utf-8`` or ``# -*- coding: utf-8 -*-`` as first line
 * Prefer `%` over `.format()`, prefer `%(varname)` instead of positional.
   This is better for translation and clarity.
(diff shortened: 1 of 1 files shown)


pedrobaeza:
OK, seems reasonable :+1:

dreispt:
:+1:

moylop260:
Can you add a comment in [differences-with-odoo-guidelines](https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#differences-with-odoo-guidelines)?

moylop260:
small comment after that :+1:

osvalr:
this is very useful and makes coding guidelines more clearer :+1:

PR #125, opened 2015-09-15
[REF] autopep8: Delete vim comment

moylop260:


tools/autopep8_extended.py
```diff
@@ -20,8 +20,29 @@ def __init__(self, pep8_options, source):
...
+        'Detect vim comment'
+        msg_code = 'CW0002'
+        check_result = []
+        msg = self.msgs[msg_code]
+        line_no = 0
+        for line in self.source:
+            column = 0
+            line_no += 1
+            line = line.strip()
+            if line.startswith('#') and \
```
pedrobaeza: Why don't use here a regular expression?

moylop260: In this case regex is a good idea... line.startwith too.
I saw a post of problems with regex recently...
I prefered avoid them in this simple case.

pedrobaeza: OK then

pedrobaeza:
:+1:

max3903:
:+1:

max3903:
@moylop260 can you fix the conflict so we can merge it ? Thank you!

moylop260:
@max3903 
rebased

PR #130, opened 2015-09-20
Added the connector-odoo2odoo repo

elicoidal:


tools/oca_projects.py.save
```diff
@@ -0,0 +1,170 @@
+# -*- coding: utf-8 -*-
```
pedrobaeza: You have to remove this backup file

elicoidal: OK: shall I remove and then commit again?

pedrobaeza: Remove it, and make:
```
git add --all
git commit --amend
```
This way, you won't have an extra commit.

elicoidal: great: done

pedrobaeza: I'm afraid it's still there. When pushing, you have to make git push -f

elicoidal: Not sure I didnot do something stupid...

elicoidal: Actually I did... I pushed on the main branch instead of mine
Sorry about that.

pedrobaeza: OK, the branch contains now correct things, so let's stay this way.

pedrobaeza:
@elicoidal, you have to push again your changes, because I can't retrieve then after restoring properly the branch. I have also protected the branch against force pushing to avoid this kind of incidents again.

elicoidal:
@pedrobaeza OK I will. Another option is to remove my super-powers.
I dont need them and I dont know how to use them...

pedrobaeza:
"A great power means a great responsability"  :wink:

No, there's no problem now because the "super-powers" also allow you to merge some branches when needed, but now the main branch is protected to avoid this by mistake.

elicoidal:
Done. Thanks for your help and patience!

moylop260:
FYI github has a new validation to avoid `git push -f` in stable branches.
no-push-f

pedrobaeza:
Thanks, @moylop260. I activated it when this came.

PR #140, opened 2015-09-25
[REM] useless link https://www.odoo.com/forum/help-1

tafaRU:
Following https://github.com/OCA/server-tools/pull/261#discussion_r40427477


pedrobaeza:
As we already have Bug tracker section, I see the other link useless (because OCA doesn't attend specifically Odoo general forum), so :+1: 

Also consider to set an OCA forum is not the more proper thing, as people will mix concepts regarding specific OCA topics with another newbie general questions.

eLBati:
:+1:

StefanRijnhart:
:+1:

moylop260:
:+1:

dreispt:
:100:

PR #141, opened 2015-10-02
[ADD] migrate_branch: Tool for migrating branches for Odoo versions

pedrobaeza:
This script helps to create a new branch for a new Odoo version from the
another existing branch, making the needed changes on contents.

Installation
============

For using this utility, you need to install these dependencies:

* github3.py library for handling Github calls. To install it, use:
  `sudo pip install github3.py`.

Configuration
=============

You must have a file called oca.cfg on the same folder of the script for
storing credentials parameters. You can generate an skeleton config running
this script for a first time.

Usage
=====
```
oca-migrate-branch [-h] [-p PROJECTS [PROJECTS ...]] [-e EMAIL]
                        [-t TARGET_ORG]
                        source target

positional arguments:
  source                Source branch (existing)
  target                Target branch (to create)

optional arguments:
  -h, --help            show this help message and exit
  -p PROJECTS [PROJECTS ...], --projects PROJECTS [PROJECTS ...]
                        List of specific projects to migrate
  -e EMAIL, --email EMAIL
                        Provides an email address used to commit on GitHub if
                        the one associated to the GitHub account is not public
  -t TARGET_ORG, --target-org TARGET_ORG
                        By default, the GitHub organization used is OCA. This 
                        arg lets you provide an alternative organization
```
This script will perform the following operations for each project:

* Create a branch starting from branch 'source' with 'target' as name. If it
  already exists, then the project is skipped.
* Mark all modules as installable = False.
* Replace in README.md all references to source branch by the target branch.
* Replace in .travis.yml all references to source branch by the target branch.
* Remove \__unported\__ dir.
* Make target branch the default branch in the repository.
* Create a milestone (if not exist) for new version.
* Create an issue enumerating the modules to migrate, with the milestone
  assigned, and with the labels "help wanted" and "work in progress" (if
  exist).

Known issues / Roadmap
======================

* Modules without installable key in the manifest are filled with this key,
  but the indentation for this added line is assumed to be 4 spaces, and the
  closing brace indentation is 0.
* Issue enumerating the module list contains a list to a Wiki page that should
  be formatted this way:
  https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-{branch}
* Make the created branch protected (no support yet from github3 library).

### DIFF
--- a/tools/migrate_branch.py
+++ b/tools/migrate_branch.py
@@ -1,0 +1,265 @@
+#!/usr/bin/env python
+#  -*- coding: utf-8 -*-
+"""
+This script helps to create a new branch for a new Odoo version from the
+another existing branch, making the needed changes on contents.
+
+Installation
+============
+
+For using this utility, you need to install these dependencies:
+
+* github3.py library for handling Github calls. To install it, use:
+  `sudo pip install github3.py`.
+
+Configuration
+=============
+
+You must have a file called oca.cfg on the same folder of the script for
+storing credentials parameters. You can generate an skeleton config running
+this script for a first time.
+
+Usage
+=====
+oca-migrate-branch [-h] [-p PROJECTS [PROJECTS ...]] [-e EMAIL]
+                        [-t TARGET_ORG]
+                        source target
+
+positional arguments:
+  source                Source branch (existing)
+  target                Target branch (to create)
+
+optional arguments:
+  -h, --help            show this help message and exit
+  -p PROJECTS [PROJECTS ...], --projects PROJECTS [PROJECTS ...]
+                        List of specific projects to migrate
+  -e EMAIL, --email EMAIL
+                        Provides an email address used to commit on GitHub if
+                        the one associated to the GitHub account is not public
+  -t TARGET_ORG, --target-org TARGET_ORG
+                        By default, translation are committed in GitHub on
+                        OCA. This arg lets you provide an alternative
+                        organization
+
+This script will perform the following operations for each project:
+
+* Create a branch starting from branch 'source' with 'target' as name. If it
+  already exists, then the project is skipped.
+* Mark all modules as installable = False.
+* Replace in README.md all references to source branch by the target branch.
+* Replace in .travis.yml all references to source branch by the target branch.
+* Make target branch the default branch in the repository.
+* Create a milestone (if not exist) for new version.
+* Create an issue enumerating the modules to migrate, with the milestone
+  assigned, and with the labels "help wanted" and "work in progress" (if
+  exist).
+
+Known issues / Roadmap
+======================
+
+* Modules without installable key in the manifest are not disabled.
+* Issue enumerating the module list contains a list to a Wiki page that should
+  be formatted this way:
+  https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-{branch}
+* Make the created branch protected (no support yet from github3 library).
+
+Credits
+=======
+
+Contributors
+------------
+
+* Pedro M. Baeza <pedro.baeza@serviciosbaeza.com>
+
+Maintainer
+----------
+
+.. image:: https://odoo-community.org/logo.png
+   :alt: Odoo Community Association
+   :target: https://odoo-community.org
+
+This module is maintained by the OCA.
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose
+mission is to support the collaborative development of Odoo features and
+promote its widespread use.
+
+To contribute to this module, please visit http://odoo-community.org.
+"""
+
+import argparse
+import re
+from . import github_login
+from . import oca_projects
+from .config import read_config
+
+
+class BranchMigrator(object):
+    def __init__(self, source, target, target_org=None, email=None):
+        # Read config
+        config = read_config()
+        self.gh_token = config.get('GitHub', 'token')
+        # Connect to GitHub
+        self.github = github_login.login()
+        gh_user = self.github.user()
+        if not gh_user.email and not email:
+            raise Exception(
+                'Email required to commit to github. Please provide one on '
+                'the command line or make the one of your github profile '
+                'public.')
+        self.gh_credentials = {'name': gh_user.name or str(gh_user),
+                               'email': gh_user.email or email}
+        self.gh_source_branch = source
+        self.gh_target_branch = target
+        self.gh_org = target_org or 'OCA'
+
+    def _replace_content(self, repo, path, replace_list):
+        # Re-read path for retrieving content
+        gh_file = repo.contents(path, self.gh_target_branch)
+        content = gh_file.decoded
+        for replace in replace_list:
+            content = re.sub(replace[0], replace[1], content, flags=re.DOTALL)
+        new_file_blob = repo.create_blob(content, encoding='utf-8')
+        return {
+            'path': path,
+            'mode': '100644',
+            'type': 'blob',
+            'sha': new_file_blob
+        }
+
+    def _create_commit(self, repo, tree_data, message):
+        if tree_data:
+            branch = repo.branch(self.gh_target_branch)
+            tree_sha = branch.commit.commit.tree.sha
+            tree = repo.create_tree(tree_data, tree_sha)
+            commit = repo.create_commit(
+                message=message, tree=tree.sha, parents=[branch.commit.sha],
+                author=self.gh_credentials, committer=self.gh_credentials)
+            repo.ref('heads/{}'.format(branch.name)).update(commit.sha)
+            return commit
+
+    def _mark_modules_uninstallable(self, repo, root_contents):
+        """Make uninstallable the existing modules in the repo."""
+        tree_data = []
+        modules = []
+        for root_content in root_contents.values():
+            if root_content.type != 'dir':
+                continue
+            module_contents = repo.contents(
+                root_content.path, self.gh_target_branch)
+            manifest = module_contents.get('__openerp__.py')
+            if manifest:
+                modules.append(root_content.path)
+                # Re-read path for retrieving content
+                tree_data.append(self._replace_content(
+                    repo, manifest.path,
+                    [('["\"]installable["\"]:[ ]*True',
+                      "'installable': False")]))
+        self._create_commit(
+            repo, tree_data, "[MIG] Make modules uninstallable")
+        return modules
+
+    def _update_metafiles(self, repo, root_contents):
+        """Update metafiles (README.md, .travis.yml...) for pointing to
+        the new branch.
+        """
+        tree_data = []
+        source_string = self.gh_source_branch.replace('.', '\.')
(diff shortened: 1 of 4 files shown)


pedrobaeza:
You can see an example in this branch: https://github.com/pedrobaeza/vertical-association

And this is the created issue: https://github.com/pedrobaeza/vertical-association/issues/2

pedrobaeza:
@sbidoul, @max3903, @legalsylvain, @gurneyalex can you please check?

pedrobaeza:
When you tell me you see it correct, I'll make the mass-update.

tools/migrate_branch.py
```diff
@@ -0,0 +1,265 @@
...
+                continue
+            module_contents = repo.contents(
+                root_content.path, self.gh_target_branch)
+            manifest = module_contents.get('__openerp__.py')
+            if manifest:
+                modules.append(root_content.path)
+                # Re-read path for retrieving content
+                tree_data.append(self._replace_content(
+                    repo, manifest.path,
+                    [('["\"]installable["\"]:[ ]*True',
```
StefanRijnhart: Don't you mean '["\']installable["\']:[ ]*True' (with a single quote after the backslash).

What happens if there is no 'installable' key (because True is the default. I always leave it out of my manifests)?

max3903:
@pedrobaeza Thanks Pedro. Does it move unported modules up as well ?

tools/migrate_branch.py
pedrobaeza: This is a known issue put on the instructions. I remember in OCA all modules this key is already assured and checked on reviews.

sbidoul: @pedrobaeza I noticed some modules don't have the 'installable' key at all in their manifest. That's ok since the default value is True. But it needs to be created in such cases.

tools/migrate_branch.py
```diff
@@ -0,0 +1,265 @@
...
+        target_string = self.gh_target_branch
+        source_string_dash = self.gh_source_branch.replace('.', '-')
+        target_string_dash = self.gh_target_branch.replace('.', '-')
+        if root_contents.get('README.md'):
+            tree_data.append(self._replace_content(
+                repo, 'README.md',
+                [(source_string, target_string),
+                 (source_string_dash, target_string_dash),
+                 ("\[//]: # \(addons\).*\[//]: # \(end addons\)",
+                  "[//]: # (addons)\n[//]: # (end addons)")]))
```
sbidoul: Instead of doing this replacement, run oca-gen-addons-table on the new branch and after setting installable=False in the manifests.

pedrobaeza:
@max3903, no, my script doesn't touch unported dir. According OCA rules, this directory should be emptied because it means that the modules haven't been migrated on 2 versions. I can try to include on the script.

tools/migrate_branch.py
```diff
@@ -0,0 +1,265 @@
...
+* Create a branch starting from branch 'source' with 'target' as name. If it
+  already exists, then the project is skipped.
+* Mark all modules as installable = False.
+* Replace in README.md all references to source branch by the target branch.
+* Replace in .travis.yml all references to source branch by the target branch.
+* Make target branch the default branch in the repository.
+* Create a milestone (if not exist) for new version.
+* Create an issue enumerating the modules to migrate, with the milestone
+  assigned, and with the labels "help wanted" and "work in progress" (if
+  exist).
```
sbidoul: This procedure seems fine and very complete. Great job!

sbidoul:
I think we can leave `__unported__` in the 9.0 branch if it is present in 8.0.

tools/migrate_branch.py
sbidoul: There are ~90 installable modules in 8.0 without the installable key.

pedrobaeza: OK, I need to find then a procedure to handle this. Let me do it.

tools/migrate_branch.py
pedrobaeza: Well, I'm dropping this part to not confuse people seeing this, but the logical thing is to run afterwards your script. Do you want me to do it in the same script?

elicoidal:
I thought we were getting rid of unported directory.

sbidoul:
Or `git mv __unported__/* . && rmdir __unported__` in the 8.0 branch *before* creating 9.0.

tools/migrate_branch.py
sbidoul: It's not necessary to do it in the same script indeed.

pedrobaeza: The problem with your script is that it's designed to work locally, and mine is writing directly to GitHub, making it extremely fast, so there's no possibility to use your script.

pedrobaeza:
I have already added the handle of manifests without installable key.

About unported dir, I cannot use git rm because I'm not downloading branches. I have explored the possibility of making the deletion of the folder via GitHub, but it's not possible for folders, only for files. It rests to explore a way to make a commit tree with all the deletions, but I think we can leave without this deletion and proceed with the migration.

max3903:
@pedrobaeza what happens if you just `git mv __unported__/* .` and leave the deletion of the directory to the first pull request on 9.0 ? Does it work ?

tools/oca_projects.py
```diff
@@ -136,12 +136,15 @@
...
         'odoo-community.org',
-        'community-data-files',
         'contribute-md-template',
-        'website',
-        ])
+        'maintainer-tools',
+        'maintainer-quality-tools',
+        'odoo-sphinx-autodoc',
+        'openupgradelib',
+        'connector-magento-php-extension',
```
max3903: OCB and OpenUpgrade?

pedrobaeza: Right. Done!

max3903:
:+1:

pedrobaeza:
OK, now the script also removes __unported__ dir, as you can see in https://github.com/pedrobaeza/vertical-association/commit/3a50c64bdd178fe417000fa4dabc3769bab4b1b2, so everything is completed

sbidoul:
@pedrobaeza @max3903 should we take the opportunity to add the 8.0 and 9.0 prefix to addons versions? Are there side effects to such a change?

I have a small script that does it and I can push the change to all repos if you want.

max3903:
@sbidoul Pedro's script could prefix with 8.0, but changing the version from 8.0.X.Y.Z to 9.0.1.0.0 should be done by the contributor who migrate the module.

pedrobaeza:
I agree with @max3903 that this version switch should be done by the contributor who makes the migration of the module. @sbidoul, when you run the gen_addons script, the list will contain 8.0 because they will be unported, so I see that correct.

I'm seeing also that you're making the update manually repo by repo, which is quite annoying. Why don't you see this script and try to adapt yours for scanning all OCA repositories and commit directly to GitHub without needing to clone anything?

sbidoul:
> changing the version from 8.0.X.Y.Z to 9.0.1.0.0 should be done by the contributor who migrate the module.

Agreed.

> I'm seeing also that you're making the update manually repo by repo, which is quite annoying. Why don't you see this script and try to adapt yours for scanning all OCA repositories and commit directly to GitHub without needing to clone anything?

Fortunately I'm not doing it manually :smile: I'm just doing it the old way with a script working on files and unix commands (due to my old age maybe). This script now needs to be put in a cron on an OCA server. Working on repo clones is not an issue to me. Or is it?

Regarding settings the version format to 8.0.x.y.z in all modules, I meant doing that in the 8.0 branch, before branching 9.0.

I also meant doing `git mv __unported__/* .` in the 8.0 branch, before branching 9.0.

@pedrobaeza @max3903 What do you think?

sbidoul:
I just made a couple of tests and so far I don't see any negative side effects in normalizing version numbers and moving uninstallable addons out of unported on the 8.0 branch. Travis seem to be happy with that and I suppose runbot too.

So the steps would go as follow:
1. normalize version numbers in the 8.0 branch (example here: https://github.com/acsone/account-financial-tools/commit/7d3b44a202c74251ad59da9f3398f8176eef27d0)
2. move content of `__unported__` to repo root in the 8.0 branch (example here: https://github.com/acsone/account-financial-tools/commit/57b1f5e1077e533af22ad38c25987c9fc9626e3a)
3. run @pedrobaeza 's script to branch 9.0

This way we'll have a clean addons table ([example](https://github.com/acsone/account-financial-tools/blob/8.0-version-unported-demo-sbi/README.md)) in the 9.0 branch showing all unported modules and showing clearly which come from 8.0 (with version 8.0.x.y.z) and which are older and unported in 8.0 (version x.y.z).

If you agree with this procedure, I can inform the contributors and then run 1. and 2. on Tuesday.

sbidoul:
Hm. On closer inspection it seems travis tests uninstallable modules too. So I made https://github.com/OCA/maintainer-quality-tools/pull/263.

tools/oca_projects.py
max3903: @pedrobaeza I just added pylint-odoo, can you add it please ?

pedrobaeza: Done

pedrobaeza:
Should I proceed with the mass-upgrade?

StefanRijnhart:
Thanks for fixing my nag. :+1: and go ahead then.

pedrobaeza:
@sbidoul, is OK also for you? My script now delete unported dir, so this one is covered, and reset addons table, so you (or i) can run the gen_addons later.

max3903:
:+1: Go!

sbidoul:
What about my suggestion: https://github.com/OCA/maintainer-tools/pull/141#issuecomment-145344707?

pedrobaeza:
Well, for me it's not critical to normalize versions, but if you want...

sbidoul:
> Well, for me it's not critical to normalize versions, but if you want...

There is also my proposal to move stuff out of unported on the 8.0 branch before branching 9.0. In that way the README on the default 9.0 branch will show all unported addons (those from 8.0 and those olders).

pedrobaeza:
But do you have that script done? If yes, go ahead. If not, I will not wait, but let's say what others say.

sbidoul:
@pedrobaeza the scripts are ready and tested. I can launch them anytime. 

Note we need to merge https://github.com/OCA/maintainer-quality-tools/pull/263 before moving stuff out of unported or branching 9.0, otherwise travis will become all red because of uninstallable modules.

pedrobaeza:
OK, let's finish the other PR first. I have asked you about the coverage part.

pedrobaeza:
I merge and proceed now with the migration of 9.0. Let's go!

StefanRijnhart:
Thank you @pedrobaeza! Checking some repos and they all look good except for https://github.com/oca/sale-workflow/tree/9.0. Have you got logs to check what was different about this one?

pedrobaeza:
Fixed! That repo remained in an intermediate state (branch 9.0 created, but no adjustment commits).

PR #148, opened 2015-10-16
[REF] contributing: Add unitest sequential name guideline

moylop260:
Unitest run `test*` method alphabetically.
With number `test_10*` and `test_20` force that ever run `test_10` first and second `test_20`

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -575,6 +575,24 @@ increased or decreased. If it has decreased, this is usually a sign that a test
 should be added. The coveralls web interface can also show which lines need
 test cases.
 
+### Unitest
+
+If you use unitest class use a sequential name in your method test.
+E.g.
+```python
+from openerp.tests.common import TransactionCase
+
+class TestUom(TransactionCase):
+    def test_10_caseA(self):
+      pass
+
+    def test_20_caseB(self):
+      pass
+
+    def test_30_caseC(self):
+      pass
+```
+
 ## Git
 
 ### Commit message


lmignon:
@moylop260 what's the motivation behind your proposal?

pedrobaeza:
Tests usually don't depend on the order, so this is not really needed.

moylop260:
There are many examples of use named sequential in methods from odoo base:
 [survey](https://github.com/odoo/odoo/blob/31ff6c68f4b49d5c12e2487d369a338c51f8997b/addons/survey/tests/test_survey.py#L45-L63)
[resource](https://github.com/odoo/odoo/blob/31ff6c68f4b49d5c12e2487d369a338c51f8997b/addons/resource/tests/test_resource.py#L78-L133)
...

I want to know your opinion of this practices

pedrobaeza:
As stated, I see it useless.

nhomar:
Hello All.

Our proposal here is because we have several cases where the order of tests matter:

If it is a TestResourceCommon or a TransactionCase.

Then the common accepted strtegy is put numbers on your tests to order them just to be sure in that order the previous set was done correctly, because may be you test without such configuration will fail.

Even the templates for python module add tests with 00X on the name in order to propose you that.

from my PoV it hase some value and give some visual order.

Then I think the guideline is good.

nhomar:
You can read a better discussion here:

http://stackoverflow.com/questions/4095319/unittest-tests-order

lmignon:
IMO it's a bad practice to rely on the test order. Each test must be state less and lets the used resources unmodified. It's common to factorize into helper functions the creation of some resources that are the central point of some cases and used to setup the required env in others cases. By writing your tests in this way you avoid to rely on the test order.

lasley:
I second this - each test should be independent from the others in order to completely isolate what is failing without making everything else go red at the same time. IMO you are able to get a much more clear picture of the problem when only 1-2 tests specifically relating to the bug fail, instead of 10-12 failing due to a dependency.

Along with the helper methods, I have found that creative use of Mocks also help by allowing you to stub out method dependency chains.

CONTRIBUTING.md
```diff
@@ -575,6 +575,24 @@ increased or decreased. If it has decreased, this is usually a sign that a test
...
 test cases.
 
+### Unitest
+
+If you use unitest class use a sequential name in your method test.
+E.g.
+```python
+from openerp.tests.common import TransactionCase
+
+class TestUom(TransactionCase):
```
Yajo: I only see this useful if you change this for `SingleTransactionCase`.

osvalr:
:+1: with this change, because certainly is not a good practice to depend upon test cases order, but it is a good practice to have all unit test cases ordered for later check and locate what failed and why, some modules have a bunch of use case in tests so this will help to reach and locate it quickly.

elicoidal:
:-1:  
Test should be independent and stateless

lasley:
My count on this is that the no's have it. Tally:

* @lmignon - N
* @pedrobaeza - N
* @lasley - N
* @nhomar - Y
* @osvalr - Y
* @elicoidal - N

That's 4 nos and 2 yes. Counting @moylop260 as submitter puts it at 3 yes.

In light of the above, I'm closing this one. Thanks for the submission @moylop260 (as always)

PR #149, opened 2015-10-22
[IMP] add --remove-old-repos options to oca-clone-everything

sbidoul:
This is a prerequisite to fully automating the oca-gen-addons-table runs.

### DIFF
--- a/tools/clone_everything.py
+++ b/tools/clone_everything.py
@@ -7,7 +7,8 @@ from oca_projects import OCA_REPOSITORY_NAMES, url
 import os
 
 
-def clone(organization_remotes=None):
+def clone(organization_remotes=None,
+          remove_old_repos=False):
     for project in OCA_REPOSITORY_NAMES:
         cmd = ['git', 'clone', '--quiet', url(project), project]
         try:
@@ -27,6 +28,12 @@ def clone(organization_remotes=None):
                 cmd = ['git', '--git-dir=' + os.path.join(project, '.git'),
                        'fetch', organization_remote]
                 subprocess.call(cmd)
+    if remove_old_repos:
+        for d in os.listdir('.'):
+            if d not in OCA_REPOSITORY_NAMES and \
+                    os.path.isdir(d) and \
+                    os.path.isdir(os.path.join(d, '.git')):
+                subprocess.check_call(['rm', '-fr', d])
 
 
 def main():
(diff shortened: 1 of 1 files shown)


gurneyalex:
the script works with github if you are authenticated already. The hardcoded list should be completely removed IMO.

pedrobaeza:
Let me work on this in the following days to avoid the need to clone everything

sbidoul:
@gurneyalex 

> The hardcoded list should be completely removed IMO.

You mean the hardcoded list in https://github.com/OCA/maintainer-tools/blob/master/tools/oca_projects.py?

> the script works with github if you are authenticated already

Yes. Do you mean it can not work with a generic OCA bot user?

sbidoul:
@pedrobaeza 

> Let me work on this in the following days to avoid the need to clone everything

Why? This script works and is well tested. Note the clone only occurs only the first time (then it's only fetching the changes).

pedrobaeza:
But you need to keep that storage room for all the repositories. Have you seen migrate_branch.sh? I only get what's relevant from the repo. For example, the repo manifest: https://github.com/OCA/maintainer-tools/blob/master/tools/migrate_branch.py#L165, and then, making the commit via the GitHub API, not with local git.

sbidoul:
@pedrobaeza I have seen migrate branch. I understand in some situations it's good to manipulate the github API directlly. In other situations it is either necessary to have a local clone of everything (for instance #156), or simply easier to works with with simple shell scripts.

Regarding this PR, is there an objection to have this new `--remove-old-repos` options?

pedrobaeza:
:+1:

gurneyalex:
@sbidoul yes that's the list I was refering to. It's a pain to maintain and the oca_projects.py only defaults to this list if you have not signed in github (using oca-github-login), otherwise the list is fetched from github. 

@pedrobaeza please don't break oca-clone-everything, we use it daily to keep a local mirror of all OCA projects and branches. And it is precious when you want to run mass commit scripts. 

:+1:

pedrobaeza:
Why I'm going to break oca-clone-everything? That's not what I want. I just want to change this script to not need this previous step.

tools/clone_everything.py
```diff
@@ -27,6 +28,12 @@ def clone(organization_remotes=None):
                 cmd = ['git', '--git-dir=' + os.path.join(project, '.git'),
                        'fetch', organization_remote]
                 subprocess.call(cmd)
+    if remove_old_repos:
+        for d in os.listdir('.'):
+            if d not in OCA_REPOSITORY_NAMES and \
+                    os.path.isdir(d) and \
+                    os.path.isdir(os.path.join(d, '.git')):
+                subprocess.check_call(['rm', '-fr', d])
```
gurneyalex: `shutil.rmtree`?

pedrobaeza:
I was referring the Bidoul's script that update the README's, which was the original reason to modify this

gurneyalex:
@pedrobaeza ok. I would not bother (unless you have lots of time). This is going to run on one of the OCA's big servers :smile_cat:

PR #164, opened 2015-11-27
[WIP] Add a section for installation hooks to define its use in contributing

osvalr:
I think this is necessary to point out how to define installation hooks methods and also having a separated file for them

The feedback is welcomed on this one


moylop260:
:+1:

nhomar:
:+1:

PR #166, opened 2015-11-28
[WIP] Add a section for installation hooks to define its use in contributing

osvalr:
I think this is necessary to point out how to define installation hooks methods and also having a separated file for them

The feedback is welcomed on this one

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -96,7 +96,8 @@ following
 }
 ```
 
-Remember to add into the **`__init__.py`** the following imports as needed
+Remember to add into the **`__init__.py`** the following imports as
+needed. For example:
 ```python
 ...
 from .hooks import pre_init_hook
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -80,12 +80,38 @@ For `static files`, the name pattern is `<module_name>.ext` (i.e.
 ...). Don't link data (image, libraries) outside Odoo: don't use an url to an
 image but copy it in our codebase instead.
 
+### Installation hooks
+When **`pre_init_hook`**, **`post_init_hook`** and **`uninstall_hook`** are
+used, they should be placed in **`hooks.py`** located at the root of module
+directory structure and keys in the manifest file keeps the same as the
+following
+
+```python
+{
+    ...
+    'pre_init_hook': 'pre_init_hook',
+    'post_init_hook': 'post_init_hook',
+    'uninstall_hook': 'uninstall_hook',
+    ...
+}
+```
+
+Remember to add into the **`__init__.py`** the following imports as needed
+```python
+...
+from .hooks import pre_init_hook
+from .hooks import post_init_hook
+from .hooks import uninstall_hook
+...
+```
+
 The complete tree should look like this:
 
 ```
 addons/<my_module_name>/
 |-- __init__.py
 |-- __openerp__.py
+|-- hooks.py
 |-- controllers/
 |   |-- __init__.py
 |   `-- main.py
@@ -740,6 +766,7 @@ The differences include:
     * Separating data and demo data xml folders
     * Not changing xml_ids while inheriting
     * Add guideline to use external dependencies
+    * Define a separated file for installation hooks
 * [XML](#xml-files)
     * Avoid use current module in xml_id
     * Use explicit `user_id` field for records of model `ir.filters`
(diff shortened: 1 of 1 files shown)


moylop260:
:+1:

CONTRIBUTING.md
```diff
@@ -80,12 +80,35 @@ For `static files`, the name pattern is `<module_name>.ext` (i.e.
 ...). Don't link data (image, libraries) outside Odoo: don't use an url to an
 image but copy it in our codebase instead.
 
+### Installation hooks
+When **`pre_init_hook`** and **`post_init_hook`** are used, they should be placed
```
pedrobaeza: You have missed here `uninstall_hook` that also exists.

CONTRIBUTING.md
```diff
@@ -80,12 +80,35 @@ For `static files`, the name pattern is `<module_name>.ext` (i.e.
...
+in **`init_hooks.py`** located at the root of module directory structure and keys
```
pedrobaeza: I'm thinking that as uninstall hook is not strictly an "init hook", we should call the file simply `hooks.py`.

guewen: > I'm thinking that as uninstall hook is not strictly an "init hook", we should call the file simply hooks.py.

I agree. Also because `init` usually refers to the installation of an addons (i.e. `--init` command line).

osvalr: Oh, okay! let me take that change done.

CONTRIBUTING.md
osvalr: sure, thanks

osvalr:
Hi @moylop260 @guewen @pedrobaeza
Suggested changes have been taken in count. Have you any comments about this? Thanks you in advance

CONTRIBUTING.md
```diff
@@ -80,12 +80,37 @@ For `static files`, the name pattern is `<module_name>.ext` (i.e.
...
+    'post_init_hook':'sample_b_hook',
+    'uninstall_hook':'sample_c_hook',
+    ...
+}
+```
+
+Remember to add into the **`__init__.py`** the following imports as needed
+```python
+...
+from .init_hooks import sample_a_hook
```
pedrobaeza: You have to change this also

moylop260:
@osvalr 
Rebase, please

CONTRIBUTING.md
```diff
@@ -80,12 +80,37 @@ For `static files`, the name pattern is `<module_name>.ext` (i.e.
...
+### Installation hooks
+When **`pre_init_hook`**, **`post_init_hook`** and **`uninstall_hook`** are used, they should be placed
+in **`hooks.py`** located at the root of module directory structure and keys
+in the manifest file keeps the same as the following
+
+```python
+{
+    ...
+    'pre_init_hook':'sample_a_hook',
+    'post_init_hook':'sample_b_hook',
```
moylop260: @osvalr 
Fix lints...
(spaces after `:`)

moylop260:
@osvalr 
We can't add multiple hooks methods in a manifest key `*_hook`.
Then we should rename the methods to `pre_init_hook`, `post_init_hook`...

This methods should have callable to other ones.

eg
```python
...
def delete_a():
    pass

def create_b():
    pass

def pre_init_hook():
    delete_a()
    create_b()
```

osvalr:
@moylop260 so, do you mean that main methods names should be called as the hook they are representing instead of `'sample_a_hook'`,`'sample_b_hook'`,.. ?

osvalr:
@moylop260 this pull have been rebased

CONTRIBUTING.md
```diff
@@ -80,12 +80,38 @@ For `static files`, the name pattern is `<module_name>.ext` (i.e.
...
+{
+    ...
+    'pre_init_hook': 'pre_init_hook',
+    'post_init_hook': 'post_init_hook',
+    'uninstall_hook': 'uninstall_hook',
+    ...
+}
+```
+
+Remember to add into the **`__init__.py`** the following imports as needed
```
dreispt: Remember to add into the **`__init__.py`** the following imports as needed. For example:
        ```python
        ...
        from . import init_hooks
        ```

osvalr: Hi @dreispt `init_hooks` was changed to `hooks`, and by the way, what about :
```python
from . import hooks.*
```

moylop260: @osvalr 
Use `*` is a bad idea.
We should use the name of `pre_init_hook`, `post_init_hook` and `uninstall_hook`
Because `import *` should be extra effects and show lint errors

moylop260:
:+1:

lmignon:
:+1:

pedrobaeza:
:+1:

PR #169, opened 2015-12-23
[FIX] Avoid problems with str vs unicode when formatting

Yajo:
When formatting exception messages, and an argument is a unicode with accents, formatting fails. This fixes that by using unicodes by default.

### DIFF
--- a/template/module/exceptions.py
+++ b/template/module/exceptions.py
@@ -2,6 +2,7 @@
 # © <YEAR(S)> <AUTHOR(S)>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
+from __future__ import unicode_literals
 from openerp import _, exceptions
 
 


pedrobaeza:
:+1:

moylop260:
:+1:

dreispt:
I guess @yajo can't hear me anymore about it  :smile: , but why is an `exceptions.py` part of the template? That pattern is not in the OCA contributing the guidelines, and it's not generally used nor agreed.
IMO until there is a general agreement that it should be recommended it shouldn0t be there.

Yajo:
It points to the easiest way of having them. If anybody needs it, it's nice not to rewrite it, don't you think?

Yajo:
Hi again, this is a 1-LOC patch to a current bug in the template. Could this be merged and leave discussion on https://github.com/OCA/maintainer-tools/pull/169#issuecomment-167750240 for another moment? :innocent:

Yajo:
There are docs for that, don't you think?

PR #170, opened 2016-01-05
FIX tx_pull

nicomacr:


max3903:
@nicomacr Can you explain what you are trying to fix?

hbrunn:
@nicomacr please fix travis' complaints, and I would also be interested in https://github.com/OCA/maintainer-tools/pull/170#issuecomment-169073535

PR #177, opened 2016-02-10
update the website adresse

fabienmorin:
The previous domain is expired, update with an address which is still accessible

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -700,7 +700,7 @@ S. Raymond has written influentially about peer review in software development:
 * Are there demo data?
 
 Further reading:
-* http://insidecoding.com/2013/01/07/code-review-guidelines/
+* https://insidecoding.wordpress.com/2013/01/07/code-review-guidelines/
 
 #### There are the following important parts in a review:
 


tafaRU:
:+1:

max3903:
:+1:

elicoidal:
screenshot from 2016-02-11 12-52-48
The new URL is not a good shot either: there is a warning for expiry.

fabienmorin:
@oca-clabot : I just sent my CLA signed

tafaRU:
> there is a warning for expiry.

@elicoidal you're right, it should be better to avoid it at all then. What do you think about it? cc @fabienmorin

dreispt:
I'm not sure that link should be there in the first place.

Yajo:
:+1:

hbrunn:
how about linking to https://web.archive.org/web/20130111111953/http://insidecoding.wordpress.com/2013/01/07/code-review-guidelines ? Awfully slow, but very bad stuff has to happen to have this link going down

fabienmorin:
not good idea IMO, so slow that it is hard to use, without css, it is web from the 80's and now the address : https://insidecoding.wordpress.com/2013/01/07/code-review-guidelines/
seems to work perfectly without any advertisement.
Regards

yvaucher:
28 days later the url is still valid. Let's fix that dead link. :+1:

PR #180, opened 2016-02-22
GitHub template

legalsylvain:
Use new github template;
Ref :
- https://help.github.com/articles/helping-people-contribute-to-your-project/
- https://github.com/blog/2111-issue-and-pull-request-templates
- Odoo settings : https://github.com/odoo/odoo/tree/9.0/.github

# issue
- remove issue template in the link in each module description ; (more simple description, and avoid "copy past errors" or error when porting modules in a new version);
- add a unique file per repository, named "ISSUE_TEMPLATE" ;

# Pull Request
- add a file named "PULL_REQUEST_TEMPLATE";

# Contributing
- add a file name "CONTRIBUTING.md" with a simple link to the OCA guideline;
It so adds a banner with "Please review the guidelines for contributing to this repository." message on each PR;

Review, suggestion, ... welcome.

Once this PR accepted, we should apply the changes in each repository.

regards.


.github/ISSUE_TEMPLATE
```diff
@@ -0,0 +1,18 @@
+### Milestone (Odoo version)
+- 
+
+### Module(s)
+- 
+- 
+
+### Step to reproduce
```
pedrobaeza: s/Step/Steps

.github/PULL_REQUEST_TEMPLATE
```diff
@@ -0,0 +1,14 @@
+### Milestone (Odoo version)
```
pedrobaeza: I see this file no real advantage, as we have multiple PR types:

* New modules
* Feature fixes
* CIs fixes
* Maintenance PRs
* ...

pedrobaeza:
Seeing the PR, we still lose some possibilities from the direct link in the README: the name of the module and the version. Does it deserve to lose this?

.github/PULL_REQUEST_TEMPLATE
legalsylvain: I don't, just a possibility.
User can remove the template if it's not relevant. But in 90% of the cases, this a new module or a feature or a fix, so, milestone and module name are relevant. Don't you think ?

pedrobaeza: Well, I don't know how this works.

I always put relevant commits messages (for example, the README.rst in new files), and if this overwrites current behavior that show these messages, I prefer not to have it, because I will need again to copy/paste these messages.

legalsylvain:
Hi @pedrobaeza, thanks for your nightly review !

Well, about this link, I don't like it so much ! ;-)
- It's quite annoying for developpers to write this link;
- it's counterproductive because we say "please check if a bug yet exists" with a link and just after we set another link to write directly a bug report, without realize this check;  
- it's error prone when we copy past module, change module name, port module in a new version, we have to pay attention to update this link;

In the issue template, there is the same information. The user just has to fill both fields. 

But well, no problem to keep it if you find this link important.

regards.

pedrobaeza:
You're right also in your considerations, so there's no ideal solution. Let's go with your approach that indeed is more comfortable  for the developers.

legalsylvain:
Idea : what about to add a direct link to the CLA in the contributing file ?

pedrobaeza:
And in the PR template also

.github/PULL_REQUEST_TEMPLATE
Yajo: For what I have seen in Odoo, commit message gets injected somewhere in the template. You can just delete all the template if you still prefer old behavior.

Another option is to only put template for issues.

elicoidal:
Besides the link in the contributing.md I would suggest to add the something similar (which is in my experience the most typical issues in basic formalism reviews):
* General Contribution guidelines: https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md
* Please follow the OCA README template:
https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst
* We recommend to include a screenshot as well to help new comers (nice to have)
* Please use simple headers in all py files following OCA templates
https://github.com/OCA/maintainer-tools/blob/master/template/module/__openerp__.py#L3

Other hot topics could be added (maybe 3 or 4), not making this file too big as there is already the main contributing.md

What do you think?

Yajo:
@elicoidal Do you mean putting all that in the PR template? If so, I think that's documentation, not template. I personally prefer to keep it short.

I'd add the `.md` extension to the files, so editors know what syntax to apply.

elicoidal:
@yajo no. not in the PR template
@legalsylvain suggested to add a simple contributing with link to the main contributing.md.
I would like to add it there.

legalsylvain:
@yajo : thanks, I renamed with .md extension.

@elicoidal   
- other point : the CLA, with the link to download it and the email ?

But for all this points, I'd prefer to set all in the main contributing file, in this repository, and just have a "pointer" in each repository. Well having directly all the list is more efficient, but it will be more maintenance work when we'll have to update this list. What do you think ?

Yajo:
Right now AFAIK there's a @oca-clabot that ensures the contributor has signed the CLA and provides the link automatically.

On the other points, whatever solution requires less maintenance will get my thumb bigger.

Yajo:
Wow, just today I was submitting a bug for https://github.com/docker/docker/ and found that [their template includes useful information between these tags: `<!-- ... -->`.](https://raw.githubusercontent.com/docker/docker/master/.github/PULL_REQUEST_TEMPLATE.md) Submitter sees it, but it disappears when submitted.
That can be a neat solution to have information useful for users, but that we don't want to see in the final result.

pedrobaeza:
Yeah, very good option indeed.

elicoidal:
@Yajo indeed not sure where to put it
Maybe we should create a get_started.md for a shorter version of the contributing.md

Yajo:
IMHO after what I just discovered, your points in https://github.com/OCA/maintainer-tools/pull/180#issuecomment-187600797 are short enough to put them into the hidden section of the PR template.

Yajo:
Another good point for that quick section would be "don't include `.pot` files"

.github/CONTRIBUTING.md
```diff
@@ -0,0 +1 @@
+[Please read here the OCA Guidelines](https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md)
```
moylop260: What about use a symbolic link to this file?

moylop260: Or Is this a example of template to other projects?

moylop260:
LGTM
What is the current status for this pr?

.github/CONTRIBUTING.md
Yajo: I think it's a sample.

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,18 @@
+### Milestone (Odoo version)
+-
```
Yajo: Shouldn't this have a blank line above?

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,14 @@
...
+- 
+
+### Module(s)
+- 
+- 
+
+### Fixes
+- Fix # .
+
+Changes proposed in this pull request:
```
Yajo: What about `### New features`?

bealdav: ### Fixes / new features

could be sufficient ?

Yajo: yes

hparfr:
I suggest to remove Milestone field from the PR template in favor of having it in the name of branch / title of the PR.

bealdav:
ok, milestone in body of PR is duplicated data from milestone in branch name. Keep it Simple, please

.github/ISSUE_TEMPLATE.md
legalsylvain: It doesn't seem to be necessary. doesn't it ? see preview : 
https://github.com/grap/maintainer-tools/blob/github_template/.github/ISSUE_TEMPLATE.md

legalsylvain:
Hi @hparfr, milestone seems to be a duplicated information but people that make pull request doesn't have the right to set labels or Milestone, so, i suggest to let this information to make more easy PSC members that will set flags.
What do you think ?

hparfr:
Let's forget the branch name, anyone could name it as he want on his own fork.

But, the title of the branch is free and can have some semantics. It can be particullary usefull on the /pulls list page.
We can leverage the tip suggested by Yajo as a reminder for "how to title your PR"


For instance: 
```md
<!-- the title of the pr should be prefixed with the major version : 
[9] module-xyz fix perf issues
-->
# Description
<!-- What your PR does and what it resolves --> 

# Modules
<!-- modules impacted by the PR -->
```

yvaucher:
As I have almost never seen this hidden template in link used, it is certainly a better idea to have a proper template for it.

:+1:  to make it simple

max3903:
:+1:

moylop260:
:+1:

PR #183, opened 2016-03-12
[REF] CONTRIBUTING.md: Add section for "don't bypass the ORM"

moylop260:
Important section extracted from: [odoo/coding_guidelines_framework](https://doc.odoo.com/contribute/15_guidelines/coding_guidelines_framework.html#do-not-bypass-the-orm)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -530,6 +530,33 @@ Before continuing, please be sure to read the online documentation of pyscopg2 t
   - [Advanced parameter types](http://initd.org/psycopg/docs/usage.html#adaptation-of-python-values-to-sql-types)
 
 
+### Do not bypass the ORM
+
+You should never use the database cursor directly when the ORM can do the same thing! By doing so you are bypassing all the ORM features, possibly the transactions, access rights and so on.
+
+And chances are that you are also making the code harder to read and probably less secure (see also next guideline):
+
+```python
+# very very wrong
+cr.execute('select id from auction_lots where auction_id in (' +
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -530,6 +530,33 @@ Before continuing, please be sure to read the online documentation of pyscopg2 t
...
+
+You should never use the database cursor directly when the ORM can do the same thing! By doing so you are bypassing all the ORM features, possibly the transactions, access rights and so on.
+
+And chances are that you are also making the code harder to read and probably less secure (see also next guideline):
+
+```python
+# very very wrong
+cr.execute('select id from auction_lots where auction_id in (' +
+           ','.join(map(str, ids)) + ') and state=%s and obj_price>0',
+           ('draft',))
```
eLBati: Assuming ```ids``` to be a list of integers, how can injection be made here?

moylop260: If you don't have a constraint that valid explicit that the value must be a list of integer we are susceptible to sql injection with concatenate.

The sql injection is explain better [here](https://github.com/vauxoo-dev/maintainer-tools/blob/f2616656bb23b8242f3858c132c1812ff40fe0da/CONTRIBUTING.md#no-sql-injection)

moylop260: @eLBati 
FYI I have created follow script to understand the real issue of sql injection:
https://github.com/vauxoo-dev/gist-vauxoo/blob/master/sql_injection.py

eLBati: @moylop260 thanks

PR #187, opened 2016-03-18
Add hooks to the module template.

Yajo:
Another :gift: for the template.

### DIFF
--- a/template/module/hooks.py
+++ b/template/module/hooks.py
@@ -1,0 +1,46 @@
+# -*- coding: utf-8 -*-
+# © <YEAR(S)> <AUTHOR(S)>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+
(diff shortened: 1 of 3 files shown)


template/module/hooks.py
```diff
@@ -0,0 +1,46 @@
+# -*- coding: utf-8 -*-
+# © <YEAR(S)> <AUTHOR(S)>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
```
pedrobaeza: Change this to the new shorter form :wink:

Yajo: Yikes, true!

pedrobaeza:
:+1: 

(coveralls has turned back to the ugly option of putting a comment with the coverage. I hate this...)

lmignon:
:+1:

osvalr:
late but :+1:

PR #188, opened 2016-03-18
Reduce license headers a bit more.

Yajo:
In JS and CSS files, we save a line, in all others we save 5 chars.


pedrobaeza:
:+1:

elicoidal:
:+1:

lmignon:
:+1:

pedrobaeza:
@Yajo, I'm now seeing that we should point to http://www.gnu.org/licenses/agpl-3.0, not http://www.gnu.org/licenses/agpl, because if some day, there's a new AGPL version, the link won't be the correct.

PR #190, opened 2016-04-10
[ADD] discourage private mails to contributors

hbrunn:
too many mails...

### DIFF
--- a/template/module/README.rst
+++ b/template/module/README.rst
@@ -68,6 +68,8 @@ Contributors
 * Firstname Lastname <email.address@example.org>
 * Second Person <second.person@example.org>
 
+Do not contact contributors directly about support or help with technical issues.
+
 Funders
 -------
 


pedrobaeza:
I would even say more. Issues are for bug, not for help. I would put that they should use community forums or mailing list.

hbrunn:
@pedrobaeza how about this?

pedrobaeza:
:+1:

moylop260:
:+1:

dreispt:
I don't totally agree on the recommendations given on how to proceed.

I believe that the Project's GitHub Issues is _the_ place to ask for help, and the README already provides a link there. A second link could the corresponding PSCs page in the OCA website - there we would find specific advice, on how to collaborate, such as the mailing list.

pedrobaeza:
@dreispt, this is because you don't receive 2 o 3 mails per day asking for free help. The statement in README is clear, and the name also: issues are for bugs, not for dumb questions.

dreispt:
@pedrobaeza I agree with adding this clarification text. 
I disagree on the advice given on how to get help.
Sending people to the Odoo general Help Forum instead of opening an issue is not a good idea.

pedrobaeza:
Well, I disagree also with that. Issues are not for that purpose. I close all days dumb questions made as issues on our repos (including unrelated questions about Odoo core). That's why we should filter out radically these questions from the issues. A general forum serves for both Odoo and OCA modules.

dreispt:
> issues are for bugs, not for dumb questions.

Yes, posting to an ML would be better.
But some PSCs don't have an ML, or there are so many MLs that attention can be fragmented.
Perhaps: a) use the Contributors ML, b) create a new Help ML, c) create an OCA Help Forum?

pedrobaeza:
These questions are not for code contributors, are for the general audience. Functional people can ask several of them, so I vote for a forum. And why a specific OCA forum? As said, the beginners don't recognize the difference about OCA and Odoo, so let's guide them to a single point: the Odoo forum.

dreispt:
Maybe I'm speaking for myself, but most OCA contributors don't dwell on the Odoo Help Forum.
And I certainly wouldn't post a question on an OCA module in the Odoo forum.
But maybe that's just me.

pedrobaeza:
Well, let's see the opinion of other people then.

lasley:
I see both sides here. My main question is whether the problem is with OCA contributors, or the users of OCA products? I'm thinking it's the users, which then makes sense to direct towards the Odoo user forum IMO

dreispt:
To be clear, I'm not blocking the PR . consider it a neutral review.
My point was just to bring to light another point of view.
If I'm alone in my POV please go ahead with the PR.

hbrunn:
How about asking the oca board to create a landing site for help in the like of https://odoo-community.org/page/bugs and point to this? That's even a technically beautiful indirection because it keeps links stable

hbrunn:
@OCA/board do you have an opinion on this?

pedrobaeza:
:+1: for me

dreispt:
@hbrunn my corcern is not about issues, it is about using the Odoo official Forum for "help with questions or problems concerning this addon".
But to be honest this discussion goes beyond the proposed change, so please go ahead with the merge.

yvaucher:
I'm also against putting a link to https://www.odoo.com/forum/help-1 or as the last resort.

I agree with @dreispt that not everyone in OCA follows the official help forum. I think that some of the key contributors might not have an account or enough points to give a comment in the Odoo official Q&A.

Plus OCA has no control over it. My opinion is that ML should be prefered for questions about OCA modules.

```
too many mails...
```
@hbrunn You made too many modules and contributed too much :grin: 
Just kidding, thanks for your great work.

Still, I wonder if that line will be read by the people who are sending direct mails to all email address they could possibly find which are more or less related to their issue or question.

zaoral:
@hbrunn wrote:
> How about asking the oca board to create a landing site for help in the like of https://odoo-community.org/page/bugs and point to this? That's even a technically beautiful indirection because it keeps links stable

I really support this idea.
I think is brilliant idea.

Best way to avoid the problem about were the help forum should be.
Oca will have the control of the forum.
We can manage issues bus and questions from community and outsiders separately.

Is Poetic: using Odoo to solve an Odoo OCA problem

lasley:
Reviving this as I woke up to a few emails that I would rather have not read. I'm for this PR 👍 

My thoughts regarding help forum - I believe Odoo has enforced the use of their forums as the end all be all of places to take your questions. They even managed to dwarf StackOverflow, so I would say we do not create a fracture by creating our own forum. We should instead direct to the Odoo forum, although I know personally I'm not going to be trolling it to answer questions.

If we are going to fracture anything, I feel we should point people to StackOverflow. That way at least we don't have to maintain anything, but we also don't have to work within what I consider to be a horrible Q&A format & have a much broader community of moderators to clean stuff up.

Regarding a general help site - that's probably a good call. This way we can have internal discussions like we are here w/out needing to worry about the link death like @hbrunn pointed out.

gurneyalex:
I'm :+1: for a "don't send the authors mails", as I tend to receive quite a few of these. 

The Odoo help forum is one of the worst place to be looking for help, and I rarely use it (I admit I even tend to filter out the results from Google pointing to it when I look for an answer). 

I'm not sure about Stack Overflow. The rules of SO are that the questions must be "programming oriented", and generic questions about "how do I achieve XXX using <software>" are OT (and I wish more questions on SO were closed, but this is a different matter). 

Mailing lists would be my favorite tools for this, but maybe my age is showing up here. And also Odoo-based mailing lists are terrible, and I wish we had proper mailman / sympa / google groups / <whatever other choice which can hardly be worse than Odoo-based mailing lists>...

dreispt:
The fast merge rule of thumb says: merge now what we agree, and leave for a later PR what we don't agree.

hbrunn:
so I change the paragraph to `Please do not contact the authors directly about support or technical issues`?

elicoidal:
:+1: on @hbrunn last proposal.
:+1: on a standard landing page

hbrunn:
@dreispt @elicoidal  @lasley done

lasley:
👍

elicoidal:
:+1: 
We should quickly find an alternative and get something similar to:

> For support, please go to page https//odoo-community.org/page/support and do not directly contact the contributors.

pedrobaeza:
Please solve conflicts

hbrunn:
@pedrobaeza done

PR #191, opened 2016-05-16
[FIX] File naming in demo, data and views fixed

yaniaular:
File naming in demo, data and views fixed according to https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#file-naming


moylop260:
👍

pedrobaeza:
The conventions are not right. You shouldn't remove `_demo` and `_view` from the name. I'll elaborate further later.

pedrobaeza:
My reasoning is that if you remove the suffix `_demo`, `_view` and so on, when searching for files, you won't recognize the kind of file unless you see where is located, which is not usually available. Also the fact of having multiple file with the same name is for me very wrong.

moylop260:
@pedrobaeza 
+1 with your last comment.
We have a crossed guideline, because:
 - We have a template with `_view`, `_demo` and so on.
But at same time:
 - We have a guideline with [file-naming](https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#file-naming) without `_view` and `_demo`.

Then I understand you that the correct way is with `_view` and `_demo` suffix.

@yaniaular 
Could you modify the guideline text instead of template module?

yvaucher:
It seems we are missing the wizard in naming convention.

Plus it doesn't state `report_` prefix for qweb templates.

I agree with this PR, the file path is important to know what kind of file we are dealing with. For exemple we don't name `controllers/controller_main_model.py` and `models/model_main_model.py` so why should we name `views/main_model_view.xml`, for me it's just redundant ?

The only cases I'm unsure what is better is for wizard for which the view is inside the wizard directory.

yvaucher:
By the way, when searching files, I don't see in which case you don't want to know the path. Because if you are searching, you want at least to know in which module is contained.

    find . -path "*/demo/*.xml"

Would do the trick.

pedrobaeza:
@yvaucher, there are several cases where you can't see easily the name of the folder, so I prefer to keep it in the file name.

yvaucher:
@pedrobaeza Please give me an example of such case. Usually you should know in which directory you are.

hbrunn:
needs a rebase. I also think @pedrobaeza's concern is a bit theoretical, so :+1: from me.

As for wizards: I use `$model.py` and `$model.xml`, makes perfect sense to me

osvalr:
:+1:

elicoidal:
:+1: with @yvaucher

moylop260:
@pedrobaeza 
What do you think about the last comment of @yvaucher ?

moylop260:
@nhomar What do you think about?

hbto:
@yaniaular @moylop260 

I agree with this PR,

I can even be more orthodox on applying the conventions,
I have been deprecating the use of wizard naming, because
is actually a model like any other one, and we can recognise wizard
by searching TransientModel,
so for me wizard folder can be gotten rid.
Exception be made for report (though same rule should be applicable)
Why this? because we want to stick to the MVC paradigma.

 As always, there is the human factor, which in some case can outweigh 
any other reasoning.

Best Regards

osvalr:
This PR needs rebase

moylop260:
> @pedrobaeza
there are several cases where you can't see easily the name of the folder, so I prefer to keep it in the file name

Do you mean `find . -name "*_view.xml" -type f` maybe?

pedrobaeza:
Yeah, for example.

lasley:
Another example is tabs on an IDE, as well as directory searches through IDEs. Basically anything with a UI, you're not going to have the folder readily available unless you have that tab focused.

jcdrubay:
While talking about file names, I wonder if contribution guidelines should not be updated for **menus to be stored in a separate file** with one file per root menu and possibly in a dedicated folder... something more like:

- addons/<my_module_name>/
  - menu
    - <root_menu_name>_menu.xml

I think this would help in finding each of the menus and deduce which action is being called by each menu. Also, it would help to organize the order/sequence of the menus.

lasley:
We do the separate menus in vertical-medical, and I would agree that it has been a real win in terms of organization.

jcdrubay:
I keep the actions in the view/wizard/report files.

There is no complexity due to dependency in the manifest file as I declare all the view/wizard/report files which include the actions before the menus.

It is not so bad to not have alphabetical order, manifest will look like this which is quite clear (at least to me):
```
    'data': [
        # Data
        'data/model_one.xml',

        # Reports
        'reports/my_report.xml',

        # Views
        'views/model_one.xml',

        # Wizards
        'wizards/my_wizard.xml',

        # Menu
        'menu/sale_menu.xml',
    ]
```

One more point, shouldn't `report/` become `reports/` to be consistent with `views`, `models` and `wizards`?

pedrobaeza:
OK, let's see what others think.

elicoidal:
> One more point, shouldn't report/ become reports/ to be consistent with views, models and wizards?

Should not we stick to Odoo convention/scaffolding?

lasley:
@elicoidal - Does Odoo SA actually have any conventions? Their code is wild west for the most part from what I can tell, it seems up to most of their authors to determine style points such as these on their own.

@pedrobaeza - Alpha order is inherently broken for data files anyways. Let's say for example, we have a data file for `res_partner` and one for `medical_patient`, which is a delegate inherit from partner. The `res_partner` data file must be listed before the `medical_patient` in the manifest in order for the `ref` in the `medical_patient` file to point to an existing record.

I'm for both of the suggestions - isolated menu and pluralizing `report` folder. Altho seems like we might be straying from scope of this PR a bit as I'm looking at it.

yvaucher:
@jcdrubay one options for menus is to have the base menu in menu.xml and menu with action in the same file as the action and the view.

I don't think we want a menu directory, menu are part of the view.

yvaucher:
And agreed with @lasley , alphabetical order can't be done as the order of loading is important for references by xmlid.

lasley:
@yvaucher - I believe the proposal is a Menu file inside of the view directory, not a directory on its own.

yvaucher:
@lasley I can read that in the example and I'm not in favor of it:

```
        # Menu
        'menu/sale_menu.xml',
```

lasley:
@yvaucher Sorry, must have misunderstood your comment. Thanks for the elaboration

elicoidal:
> @elicoidal - Does Odoo SA actually have any conventions? Their code is wild west for the most part from what I can tell, it seems up to most of their authors to determine style points such as these on their own. They can't even seem to consistently decide if they want their data files all in the same folder or not.

Indeed the documentation is quite poor on the subject. Except the scaffolding, anything on directory structure (and nothing about the reports...)

jcdrubay:
I would not mind if it was in the folder views instead of a dedicated menu folder. This would perfectly do too:

```
        # Views
        'views/sale_menu.xml',
```

But the main ideas are:
- all menus record in a dedicated file `xxx_menu.xml`. This is especially useful when working on "big modules" with plenty of menus.
- action record stay in the `..._view.xml` files

elicoidal:
> I would not mind if it was in the folder views instead of a dedicated menu folder

The same for the reports.

Anyway, directory or naming convention would be much better than the current untidy situation...

lasley:
> The same for the reports.

I haven't written many reports, so forgive my newb question here, but aren't there typically `.py` files with reports too - similar to wizards?

jcdrubay:
I vote for :-1: because I don't think one way or the other makes a big difference and because the existing naming conventions seem to be more consistent with odoo/odoo source code.

pedrobaeza:
Outdated, so closing.

PR #192, opened 2016-05-17
[REF] never-commit guideline: Add a valid commit

moylop260:
### Milestone (Odoo version)
- >=8.0

### Module(s)
- (guidelines)

### Fixes / new features
- Add a valid commit way


hbrunn:
@moylop260 I think for new guidelines, we shouldn't refer to the standard (as opposed the legacy) api as *new* by now. I noticed lately that a lot of people who started with 8 as their first version get quite confused when I keep talking about the new api, which is the normal stuff from their point of view.

moylop260:
@hbrunn
You have a good point.
What is `the new` and what is `the old`? <- It is confuse for people who started

Changed!

osvalr:
:+1: for me

hbrunn:
thanks, :+1:

PR #194, opened 2016-05-27
[IMP] Template for v9 js module.

Yajo:
### Milestone (Odoo version)
- 9

### Module(s)
- template

### Fixes / new features

v9 introduces a completely new module system, and v9 tours get much improved.

Here is a sample tour file that also serves as a sample v9 module. Let's see if now we see more phantomjs tests.

@Tecnativa

### DIFF
--- a/template/module/static/src/js/module_name.tour.js
+++ b/template/module/static/src/js/module_name.tour.js
@@ -20,6 +20,7 @@ Tour.register({
             title: _t("The title of this step, appears in logs"),
             element: "div.jquery-selector-of-element-to-click-or-use",
             sampleText: "Value to enter if it is an input",
+            wait: 3000, // Milliseconds to wait
             waitFor: ".start-when-this-jquery-selector-matches-something",
             waitNot: ".start-when-this-jquery-selector-matches-nothing",
             onend: function() {
--- a/template/module/static/src/js/module_name.tour.js
+++ b/template/module/static/src/js/module_name.tour.js
@@ -4,6 +4,7 @@
 "use strict";
 odoo.define("module_name.tour", function (require) {
 
+// Dependencies here by alphabetic order. Template only for Odoo 9+.
 var Core = require("web.core");
 var Tour = require("web.Tour");
 var _t = Core._t;
--- a/template/module/static/src/js/module_name.tour.js
+++ b/template/module/static/src/js/module_name.tour.js
@@ -1,0 +1,53 @@
+/* © <YEAR(S)> <AUTHOR(S)>
+ * License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). */
+
+"use strict";
+odoo.define("module_name.tour", function (require) {
+
+var Core = require("web.core");
+var Tour = require("web.Tour");
+var _t = Core._t;
+
+
+Tour.register({
+    id: "test_module_name",
+    name: _t("Try to demostrate how to create a tour"),
+    path: "/path/where/tour/starts",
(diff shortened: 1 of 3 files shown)


pedrobaeza:
You should explain through comments that this is only for v9, as this template is used also for v8 (there's no branching here).

template/module/tests/test_something.py
```diff
@@ -26,3 +26,18 @@ def test_something(self):
         Any method starting with ``test_`` will be tested.
         """
         pass
+
+
+class UICase(HttpCase):
```
pedrobaeza: Is this only for v9 too?

template/module/views/assets.xml
```diff
@@ -27,5 +23,21 @@
...
+<template id="assets_frontend" inherit_id="website.assets_frontend">
+    <xpath expr=".">
+        <link rel="stylesheet"
+              href="/module/static/src/css/module_name.css"/>
+        <script type="text/javascript"
+                src="/module/static/src/js/module_name.js"/>
+    </xpath>
+</template>
+
+<template id="assets_editor" inherit_id="website.assets_editor">
```
pedrobaeza: Only for v9?

template/module/tests/test_something.py
Yajo: Yes and not. HttpCase has been around for awhile, but the code to execute was different before (because it's js, that it has been changed).

template/module/views/assets.xml
Yajo: not

template/module/tests/test_something.py
pedrobaeza: OK

template/module/static/src/js/module_name.tour.js
```diff
@@ -0,0 +1,54 @@
...
+    name: _t("Try to demostrate how to create a tour"),
+    path: "/path/where/tour/starts",
+    mode: "test",  // Or "tutorial" if it is intended to be used by humans
+    steps: [
+        {
+            title: _t("The title of this step, appears in logs"),
+            element: "div.jquery-selector-of-element-to-click-or-use",
+            sampleText: "Value to enter if it is an input",
+            waitFor: ".start-when-this-jquery-selector-matches-something",
+            waitNot: ".start-when-this-jquery-selector-matches-nothing",
```
hugho-ad: Can we add the option `wait`?

template/module/static/src/js/module_name.tour.js
```diff
@@ -0,0 +1,54 @@
+/* © <YEAR(S)> <AUTHOR(S)>
```
oscarolar: This is great, what about an inheritance template? https://github.com/odoo/odoo/blob/9.0/addons/website_sale_options/static/src/js/website_sale.test.js#L9

template/module/static/src/js/module_name.tour.js
oscarolar: 👍 V9 supports `wait`, very useful when testing asynchronous JS.

template/module/static/src/js/module_name.tour.js
Yajo: Future PR?

Yajo:
All is fixed, thubs up? :raised_hands:

moylop260:
Thanks @Yajo 
👍

hugho-ad:
:+1:

pedrobaeza:
:+1:

PR #195, opened 2016-05-28
[FIX] bdist_all_wheels: clean more aggressively between runs

sbidoul:
This bug had the nasty side effect of doing some crossover between
the 8.0 and 9.0 wheels on wheelhouse.odoo-community.org :(


sbidoul:
This fix is installed and running fine on wheelhouse.odoo-community.org

max3903:
:+1:

lmignon:
:+1:

PR #196, opened 2016-06-02
[REF] template-module: Add post_load hook

moylop260:
- [post_load info from odoo/odoo](https://github.com/odoo/odoo/blob/2779d463e80eaf903e789c29991a1faef8d07445/openerp/modules/module.py#L317-L322)

### DIFF
--- a/template/module/hooks.py
+++ b/template/module/hooks.py
@@ -51,5 +51,9 @@ def post_load():
 
     This is ok as the post-load hook is for server-wide
     (instead of registry-specific) functionalities.
+
+    This is very useful to create monkey patches for odoo.
+
+    Note: You do not have access to database cursor here.
     """
     raise NotImplementedError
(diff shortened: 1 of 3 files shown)


template/module/hooks.py
```diff
@@ -44,3 +44,12 @@ def uninstall_hook(cr, registry):
         Database registry, using v7 api.
     """
     raise NotImplementedError
+
+
+def post_load():
+    """Loaded before any model or data has been initialized.
+
+    This is ok as the post-load hook is for server-wide
```
pedrobaeza: Maybe you can document here for what is this useful

moylop260: @pedrobaeza 
Thanks for feedback.
Added just one useful case.
(Sorry, I don't know all useful cases)

lmignon:
:+1: Thank you @moylop260

pedrobaeza:
:+1:

PR #197, opened 2016-06-06
spell Copyright instead of symbol

lepistone:
In http://www.gnu.org/licenses/gpl-howto.html I read the sentence

    Always use the English word “Copyright”; by international convention,
    this is used worldwide, even for material in other languages. The
    copyright symbol “©” can be included if you wish (and your character set
    supports it), but it's not necessary. There is no legal significance to
    using the three-character sequence “(C)”, although it does no harm.

I think that means that the exact word "Copyright" is always necessary,
while the symbol is an optional addition.

### DIFF
--- a/template/module/__init__.py
+++ b/template/module/__init__.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# © <YEAR(S)> <AUTHOR(S)>
+# Copyright <YEAR(S)> <AUTHOR(S)>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from . import models, wizards, controllers, report
--- a/template/module/__openerp__.py
+++ b/template/module/__openerp__.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# © <YEAR(S)> <AUTHOR(S)>
+# Copyright <YEAR(S)> <AUTHOR(S)>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 {
     "name": "Module name",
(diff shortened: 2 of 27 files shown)


elicoidal:
:+1:

dreispt:
:+1:

sbidoul:
:+1:

pedrobaeza:
Reading that, I understand that the symbol is an option (not complementary), and I prefer the symbol, because it gives more to put the rest of the text. For me, :-1:

elicoidal:
@pedrobaeza Symbol is an abbreviation of Copyright and I am not sure about the rationale for keeping both.
It is like saying USA United States of America.
Here more does not mean better ("copyright" word is good enough to state that this is a copyright notice).
So still :+1: for @lepistone recommendation

pedrobaeza:
@elicoidal, I don't want to use both. I want to use only the symbol.

elicoidal:
> Always use the English word “Copyright”

If you want the symbol you need to add it on top of the word (AFAIU)

pedrobaeza:
@elicoidal, if you read the rest, it mentions as optional the copyright. I interpret it as optional replacement, not as complementary.

elicoidal:
@pedrobaeza This is crystal clear in the description:
Word "Copyright" is mandatory (**Always** use the English word “Copyright”)
Symbol is possible but optional (The copyright symbol “©” **can be** included if you wish)

I dont understand your standpoint here.

pedrobaeza:
I don't see that clear, but anyways, proceed as you want.

leemannd:
:+1:

moylop260:
👍

moylop260:
> @pedrobaeza  I don't see that clear, but anyways, proceed as you want.

I understand this as a lift of 👎
Then merging...

PR #199, opened 2016-06-22
[FIX] CONTRIBUTING.md: Model attibute order.

carlosdauden:
You can't use a function before that it's defined


pedrobaeza:
Thanks :+1:

hbrunn:
sure! :+1:

moylop260:
Just for the log history...

```python
     name = fields.Char(string='Name', default='_default_name')

     def _default_name(self):
         pass
```
Notice that the value of parameter `default` is a string.

PR #201, opened 2016-07-07
[IMP] Include tests folders in CONTRIBUTING.md

pedrobaeza:
Closes #103

PR #202, opened 2016-07-07
Odoo guideline is a tests direcory for py and yml

dreispt:
Fix tests directory guideline, following discussion on #201

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -160,11 +160,10 @@ addons/<my_module_name>/
 |       |   `-- <my_module_name>.less
 |       `-- xml/
 |           `-- <my_module_name>.xml
-|-- test/
-|   |-- <test_file>.yml
 |-- tests/
 |   |-- __init__.py
 |   |-- <test_file>.py
+|   |-- <test_file>.yml
 |-- views/
 |   |-- <main_model>.xml
 |   `-- <inherited_main_model>_views.xml


pedrobaeza:
:+1:

moylop260:
Is this really valid in the practice?
I saw test and tests folders from master version of odoo/odoo yet.

moylop260:
I mean, If you run:

`find addons/*/test/ -name *.yml | wc -l`
Result: 57

`find addons/*/tests/ -name *.yml | wc -l`
Result: 1

screen shot 2016-07-07 at 7 35 40 am

dreispt:
@moylop It's in the guidelines.

moylop260:
Could you add a `yml` dummy file in [template/module/tests](https://github.com/OCA/maintainer-tools/tree/master/template/module/tests) please?

dreispt:
@moylop260 I don't see the point. Anyway, if people feel it's necessary in should go in a different PR.

moylop260:
@dreispt Ok, I mean, the point is:

We have https://github.com/OCA/maintainer-tools/tree/master/template/module
screen shot 2016-07-07 at 8 44 58 am

We have the complete tree example
screen shot 2016-07-07 at 8 44 31 am

IMHO If we change the `template/module` we should change the text of the `complete tree` too.
If we change the `complete tree` we should change the `template/module` too.

This point is in order to avoid have a discuss similar to https://github.com/OCA/maintainer-tools/pull/191 in the future, you can see a example where the `template/module` tell us a thing and the contribute's document tell us other thing.

IMHO If we add/change the `yml` file in the contribute's document then we need concordant in the template module yet.

dreispt:
The template module is a just template, not a guideline.
If a guideline is not clear, it should be clarified in the guideline document.
In this case, py are generally considered superior to yml, so I see no point in adding a yml test template.

moylop260:
> In this case, py are generally considered superior to yml, so I see no point in adding a yml test template.

Good point
Thanks for clarify

Maybe our guideline should be not use .yml

dreispt:
> Maybe our guideline should be not use .yml

IMO we should suggest avoiding them, but nor forbid.

moylop260:
👍

osvalr:
:+1:

elicoidal:
:+1:

PR #205, opened 2016-07-15
[FIX] `use strict` inside functions.

Yajo:
### Milestone (Odoo version)
- any

### Module(s)
- module template

### Fixes / new features

Odoo automatically minifies and concatenates all scripts. Given [`use strict` may only be used at the beginning of a script or of a function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode), minified version would not be strict unless it is declared inside the function. Even worse, a strict script loaded as the 1st one could affect other non-strict ones loaded later.

So, in short, this should be considered a guideline.

@Tecnativa


carlosdauden:
:+1:

oscarolar:
:+1:

pedrobaeza:
:+1:

PR #206, opened 2016-07-20
[REF] CONTRIBUTING: Use replace is not recommended

moylop260:
**New guideline to avoid use `position=replace`**

Use of `<... position="replace">` is not recommended because
could show the error `Element ... cannot be located in parent view`
from other views inheriting with this field.

if you need use this option, must have an explicit comment
explaining why they are absolutely necessary and use a
high value of priority to avoid the error.

```xml
<record id="view_id" model="ir.ui.view">
    <field name="name">view.name</field>
    <field name="model">object_name</field>
    <field name="priority">110</field> <!--Priority greather than 100-->
    <field name="arch" type="xml">
        <!--Necessary because...-->
        <xpath expr="//field[@name='my_field_1']" position="replace"/>
    </field>
</record>
```

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -306,20 +306,20 @@ same name.
 
 Use of `<... position="replace">` is not recommended because
 could show the error `Element ... cannot be located in parent view`
-from other views inheriting with this field.
+from other inherited views with this field.
 
-if you need to use this option, must have an explicit comment
-explaining why they are absolutely necessary and use a
-high value of priority to avoid the error.
+If you need to use this option, it must have an explicit comment
+explaining why it is absolutely necessary and also use a
+high value in its `priority` (greater than 100 is recommended) to avoid the error.
 
 
 ```xml
 <record id="view_id" model="ir.ui.view">
     <field name="name">view.name</field>
     <field name="model">object_name</field>
-    <field name="priority">110</field> <!--Priority greather than 100-->
+    <field name="priority">110</field> <!--Priority greater than 100-->
     <field name="arch" type="xml">
-        <!--Necessary because...-->
+        <!-- It is necessary because...-->
         <xpath expr="//field[@name='my_field_1']" position="replace"/>
     </field>
 </record>
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -191,7 +191,6 @@ When declaring a record in XML:
 * For field declarations, the `name` attribute is first. Then place the `value`
   either in the `field` tag, either in the `eval` attribute, and finally other
   attributes (widget, options, ...) ordered by importance.
-
 * Try to group the records by model. In case of dependencies between
   action/menu/views, the convention may not be applicable.
 * Use naming convention defined at the next point
@@ -303,6 +302,29 @@ same name.
     ...
 </record>
 ```
+
+
+Use of `<... position="replace">` is not recommended because
+could show the error `Element ... cannot be located in parent view`
+from other views inheriting with this field.
+
+if you need use this option, must have an explicit comment
+explaining why they are absolutely necessary and use a
+high value of priority to avoid the error.
+
+
+```xml
+<record id="view_id" model="ir.ui.view">
+    <field name="name">view.name</field>
+    <field name="model">object_name</field>
+    <field name="priority">110</field> <!--Priority greather than 100-->
+    <field name="arch" type="xml">
+        <!--Necessary because...-->
+        <xpath expr="//field[@name='my_field_1']" position="replace"/>
+    </field>
+</record>
+```
+
 
 ### External dependencies
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -308,7 +308,7 @@ Use of `<... position="replace">` is not recommended because
 could show the error `Element ... cannot be located in parent view`
 from other views inheriting with this field.
 
-if you need use this option, must have an explicit comment
+if you need to use this option, must have an explicit comment
 explaining why they are absolutely necessary and use a
 high value of priority to avoid the error.
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -325,6 +325,8 @@ high value of priority to avoid the error.
 </record>
 ```
 
+Also, we can hide an element from the view using `invisible="1"`.
+
 
 ### External dependencies
 


lasley:
I'm good with this, as I am currently trying to salvage my website templates due to exactly this mistake. 👍 

Question - is this going to be integrated into our CIs at all, or just something to watch out for in reviews?

moylop260:
I'll work to migrate the check https://github.com/Vauxoo/pylint-odoo/commit/63524d274e89fab63782ccafd2f6aaef43cb34e2 after merge the rule.

CONTRIBUTING.md
```diff
@@ -304,6 +303,29 @@ same name.
 </record>
 ```
 
+
+Use of `<... position="replace">` is not recommended because
+could show the error `Element ... cannot be located in parent view`
+from other views inheriting with this field.
+
+if you need use this option, must have an explicit comment
```
pedrobaeza: s/need use/need to use

pedrobaeza:
We can recommend 2 options:

- Use `invisible="1"` for hiding a field if we want to "remove" an element from the view.
- Add also a related field to the hidden field for placing it in another place.

moylop260:
@pedrobaeza Thanks for feedback.
 - Fix typo
 - Added extra option `invisible=1`

zaoral:
And what do you thin about when we use `replace` in order to relocate field the field in the view. We first remove the field from the view and then add it again in another more important section of the view that matters to us?

moylop260:
> - Add also a related field to the hidden field for placing it in another place.

IMHO this could be your other PR, because this alternative solution has some problems offtopic of this PR.

moylop260:
@zaoral

In this case, you could use a explicit comment of why this replace is needed.
Add a high priority to avoid a `cannot be located in parent view`

pedrobaeza:
@zaoral @moylop260 for that case my suggestion was:
- Put invisible="1" in the original field.
- Add a related field to the original one.
- Add the related field in the view where you want.

This way, you won't find surprises when other views place elements using your moved field as reference.

zaoral:
@pedrobaeza @moylop260 

IMHO I have to say that I do not like neither of the options but I do not found any alternative by this moment.

- using related fields: I don't like to create a related field only to show the same field in another place of the view. For example I want to relocate 5 fields. This could change a simple model into a really messy model.
- using high priority: I also don't like the idea of the high priority since this could affect in the way that/what view will be displayed. 

I think that as @moylop260 mention, this should be treat into a new issue to ask for another proposals for the commmity and be widely discussed to take the better solution

zaoral:
As last resource we can use  
```xml
<field name="mode">primary</field>
```

This simulate the "copy" of the view and let us to edit it as much we like

moylop260:
@pedrobaeza 
Using a related field we are overloading the database of dummy fields for a UI matter.

You won't find surprises when other views place elements using your moved field as reference if your use `priority`. This field have this focus.

If you don't use `priority` then you will need a `related`, but if you use `priority` you don't need a `related`, and let the database cleaner and faster.

> @zaoral using high priority ... could affect in the way that/what view will be displayed.

If you don't use `priority` field then you could affect in the way that/what view will be displayed.
Because a random value is used (default priority plus id) to order of the fields replaced and displayed then could affect in the way that the view will be displayed and you will see a ugly error.

moylop260:
> @zaoral 
As last resource we can use  
```xml
<field name="mode">primary</field>
```
This simulate the "copy" of the view and let us to edit it as much we like

We are talking about inherit views here.
If you need a "copy" then is offtopic here, because you aren't using a inherit really.
Besides, you will need maintain 2 or more full views instead of just one inheriting.

moylop260:
For a better way to understad what I say I have created a [Proof Of Concept document](https://docs.google.com/document/d/1u9YiZ_rrDvgV74Hp2TftHHT83igwwHfE_UEry17oyI8/edit)

CONTRIBUTING.md
```diff
@@ -304,6 +303,31 @@ same name.
 </record>
 ```
 
+
+Use of `<... position="replace">` is not recommended because
+could show the error `Element ... cannot be located in parent view`
+from other views inheriting with this field.
+
+if you need to use this option, must have an explicit comment
```
osvalr: "If ..."
"it must have"

CONTRIBUTING.md
```diff
@@ -304,6 +303,31 @@ same name.
...
+explaining why they are absolutely necessary and use a
```
osvalr: "why it is absolutely necessary and also use a..."

pedrobaeza:
Related non stored fields don't overload the database. Of course my first option is to not move any field! You should see my talk about maintainability in Spanish Odoo days: https://odoospain.odoo.com/event/jornadas-odoo-2016-2016-06-16-2016-06-17-4/track/como-realizar-un-modulo-de-odoo-compatible-con-todo-152

CONTRIBUTING.md
```diff
@@ -304,6 +303,31 @@ same name.
...
+high value of priority to avoid the error.
```
osvalr: [Why >100 is okay](https://github.com/OCA/maintainer-tools/pull/206/files#diff-6a3371457528722a734f3c51d9238c13R320)?, in any case, I think is would be a good idea to put in it.
> " high value in its `priority` (greater than 100 is recommended) to avoid the error."

CONTRIBUTING.md
```diff
@@ -304,6 +303,31 @@ same name.
...
+
+
+```xml
+<record id="view_id" model="ir.ui.view">
+    <field name="name">view.name</field>
+    <field name="model">object_name</field>
+    <field name="priority">110</field> <!--Priority greather than 100-->
```
osvalr: greater

moylop260:
> Related non stored fields don't overload the database.

You are right
If you use fields function or related without `store=True` then in each visit to the view you are re-calculating the fields by each record.
Then now you migrate the overload the database to overload the ORM.

I don't understand in that cases we should use a related instead of a `priority`.
Could you help us with a example or POC like as my document?

moylop260:
@osvalr 
Thanks for typo feedback.
(Could you change them directly with your user?)

moylop260:
@pedrobaeza 

You have a good talk and I'm creating a guideline aligned with your talk.

You could proposal in other PR using related fields... (in fact, I don't understand why you don't created a pr to this project with your slide before...)

I we can discuss it.
IMHO this PR is good with the current content.

In the minute 23m34s of the video https://youtu.be/Hq9IbN7F9cg?t=23m34s

You have the slide:
screen shot 2016-07-20 at 2 57 20 pm

That error is valid if you don't use `priority`
Maybe I'm wrong, but if you have a POC with the real fail using it then you could added it to your new PR to justify the change.

What do you think?

pedrobaeza:
OK, I'll do it when I have some time. I'm very struggled right now.

moylop260:
Thanks @pedrobaeza 
Maybe, if you are agree with me... I can create the new PR with `related` change (after merge this one) and I expose your comments and my comments and we can wait other contributors comments...

pedrobaeza:
No problem. Let's continue with this one as is and we'll open the other one later.

osvalr:
:+1: for me

hbrunn:
:+1: for this change. The related fields thing sounds not very desirable to me. I often pondered about a
```
<myelement position="move">
  <location>xpath expression</location>
</myelement>
```
module that allows us to grab an element the way it is and move it somewhere else. This still can break inheriting views if they use a very specific xpath themselves, but will solve many of the problems in this class.

moylop260:
We have Lasley, hbrunn and osvalr approval
Could we merge it @pedrobaeza?

PR #207, opened 2016-07-26
Add `onload()` and fix `onend()` docs.

Yajo:
### Milestone (Odoo version)
- 9.0

### Module(s)
- template

### Fixes / new features
- `onend` docs were wrong.
- `onload` was not there.

@Tecnativa

### DIFF
--- a/template/module/static/src/js/module_name.tour.js
+++ b/template/module/static/src/js/module_name.tour.js
@@ -23,9 +23,13 @@ Tour.register({
             wait: 3000, // Milliseconds to wait
             waitFor: ".start-when-this-jquery-selector-matches-something",
             waitNot: ".start-when-this-jquery-selector-matches-nothing",
-            onend: function() {
-                // Custom code that should return true for step to success
-                return true;
+
+            // Triggers that could return a step title to jump to it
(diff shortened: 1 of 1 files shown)


pedrobaeza:
:+1:

template/module/static/src/js/module_name.tour.js
```diff
@@ -23,9 +23,13 @@ Tour.register({
             wait: 3000, // Milliseconds to wait
             waitFor: ".start-when-this-jquery-selector-matches-something",
             waitNot: ".start-when-this-jquery-selector-matches-nothing",
-            onend: function() {
-                // Custom code that should return true for step to success
-                return true;
+
+            // Triggers that could return a step title to jump to it
```
oscarolar: @yajo thanks a lot!,

Isn't `onload` way more versatile?

I mean, you are documenting great, doc you added must remain but I think we should change the statement for something like: `Executes any JS code,  could return a step title to jump to it.`

Thanks again.

Yajo: Well, they are js functions, I think that the part that "executes code" is quite obvious.

PR #208, opened 2016-07-28
[MIG][9.0] template: Upgrade JS API

lasley:
### Milestone (Odoo version)
- 
  v9
### Module(s)
- 
  template
### Fixes / new features
- Upgrade JS to new API
- Upgrade onload method to use core bus web_client_ready
- Add an exported method

REF https://github.com/OCA/web/pull/386#discussion_r72024916

cc @antespi @pedrobaeza

### DIFF
--- a/template/module/static/src/js/module_name.js
+++ b/template/module/static/src/js/module_name.js
@@ -1,7 +1,0 @@
-/* Copyright <YEAR(S)> <AUTHOR(S)>
- * License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). */
-
-(function ($) {
-    "use strict";
-    // Script that will be loaded when document is ready
-})(jQuery);
--- a/template/module/static/src/js/web_module_name.js
+++ b/template/module/static/src/js/web_module_name.js
@@ -1,0 +1,20 @@
+/* Copyright <YEAR(S)> <AUTHOR(S)>
+ * License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). */
+
+odoo.define('web_module_name.objectName', function(require){
+    "use strict";
+
+    var core = require('web.core');
+
+    core.bus.on('web_client_ready', null, function () {
+        // Script that will be loaded when document is ready
+    });
+
+    function methodToExport () {
+        // This method will be exported as
+        // require('module_name.object_name').methodToExport
+    }
+
+    return {methodToExport: methodToExport};
+
+});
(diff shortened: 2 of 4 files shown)


pedrobaeza:
Thanks for this update

:+1:

antespi:
:+1: Thanks @lasley

pedrobaeza:
@moylop260, there's a strange error in Travis. Do you have an idea of it?

moylop260:
I saw the same error from stable branch build
https://travis-ci.org/OCA/maintainer-tools/builds/147765216

It is a traceback from flake8

I'll use travis2docker to debug and fix it..

moylop260:
FYI I can reproduce it with t2d using the travis original image (with other environment I can't reproduce it)

The version with problem flake8 3.0.2
screen shot 2016-07-29 at 9 27 07 pm

I install a old version 2.6.2 and work fine!
screen shot 2016-07-29 at 9 29 05 pm

I will create a new PR with a freeze of version

template/module/static/src/js/module_name.js
```diff
@@ -1,7 +1,20 @@
...
+odoo.define('module_name.object_name', function(require){
     "use strict";
-    // Script that will be loaded when document is ready
-})(jQuery);
+
+    var core = require('web.core');
+
+    core.bus.on('web_client_ready', null, function () {
+        // Script that will be loaded when document is ready
+    });
```
obulkin: This approach doesn't work for `website`. Instead, the proper way to determine that the document is ready seems to be via `ready()` in `web_editor.base`, which returns a Promise object. This means that for `website`, the example here should look more like:

```
var base = require('web_editor.base');
base.ready().done(function() {
    // Script that will be loaded when document is ready
});
```

pedrobaeza: This is not used in website, but on web client (as stated). Maybe we should include both in 2 files: one for web, and another for website.

lasley: I see no reason to not include as many examples as possible. @pedrobaeza you're thinking a separate file would be good here instead of just putting both here and commenting the difference?

pedrobaeza: Yes, I'm thinking about 2 files, naming then `web_module_name.js` and `website_module_name.js`

lasley:
Split website sample into separate file and added a web_ namespace to the web ones. 

I did not add a website tour file, because I am not quite sure of the intricacies between the back and front end tours (if any). IMO it would be bad for me to add a sample I am not 100% positive of 😄

gurneyalex:
:+1:

lasley:
Squashed. This is good for merge?

PR #214, opened 2016-08-18
[FIX] test_something: Fix phantom string for backend

lasley:
### Milestone (Odoo version)
- 9.0

### Module(s)
- Http test case for backend

### Fixes / new features

* `module` is the argument to define a module to test in backend, not `mod` in v9 - not sure about other versions


rvalyi:
Well spotted. I checked on runbot, in 8.0, 9.0 and master it is module and not mod. In 7.0 it seems none of these URL schemes trigger tests but who cares about phantom in 70...
So :+1: for me

pedrobaeza:
:+1:

gurneyalex:
:+1:

PR #215, opened 2016-08-18
[ADD] template: PhantomJS Test Sample

lasley:
### Milestone (Odoo version)
- 9.0
### Module(s)
- Template
### Fixes / new features
- Adds an example for QUnit test suite injection via QWeb
- Adds PhantomJS test example

### DIFF
--- a/template/module/static/tests/js/module_name.js
+++ b/template/module/static/tests/js/module_name.js
@@ -1,0 +1,19 @@
+/* Copyright <YEAR(S)> <AUTHOR(S)>
+ * License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). */
+
+odoo.define_section('module_name', ['module_name.ExportedObject'], function(test) {
+    "use strict";
+
+    test('It should demonstrate a PhantomJS test',
+        function(assert, ExportedObject) {
+            var expect = 'Expected Return',
+                result = new ExportedObject();
+            assertStrictEqual(
+                result,
+                expect,
+                "Result !== Expect and the test failed with this message"
+            );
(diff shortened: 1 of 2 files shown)


pedrobaeza:
I think we should have branches 8.0, 9.0 of MT to accommodate all of these changes to all versions. Can you please bring the topic to mailing list?

About this, isn't needed a Python file to launch the test?

lasley:
@pedrobaeza you are correct that a Python file is also required - we already had that part in our template for some reason. 

Writing the email now

template/module/static/tests/js/module_name.js
```diff
@@ -0,0 +1,19 @@
+/* Copyright <YEAR(S)> <AUTHOR(S)>
+ * License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). */
+
+odoo.define_section('module_name', ['module_name.ExportedObject'], function(test) {
+    "use strict";
+
+    test('It should demonstrate a PhantomJS test',
```
Yajo: Please specify that this is for backend tests.

template/module/static/tests/js/module_name.js
```diff
@@ -0,0 +1,19 @@
...
+        function(assert, ExportedObject) {
+            var expect = 'Expected Return',
+                result = new ExportedObject();
+            assertStrictEqual(
```
Yajo: this must be `assert.assertStrictEqual`, right?

lasley:
Thanks @Yajo - updated

oscarolar:
👍

PR #216, opened 2016-08-20
[REF] template: Consider 'examples' folder

moylop260:
Many times we need add `examples` files.
- They are used like as reference or they are used to process out of manifest file.

A example is the following file:
 - https://github.com/OCA/l10n-switzerland/blob/8.0/l10n_ch_account_statement_base_import/data/UBS_export.csv

Inspired by https://github.com/OCA/pylint-odoo/issues/47#issue-164800453

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -113,6 +113,8 @@ A module is organized in a few directories:
   `lib/`, ...
 * `views/`: contains the views and templates, and QWeb report print templates
 * `wizards/`: wizard model and views
+* `examples/`: external files
+
 
 ### File naming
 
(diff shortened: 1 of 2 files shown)


pedrobaeza:
:+1:

osvalr:
I am :+1: for the solution given to the related issue

gurneyalex:
:+1:

CONTRIBUTING.md
```diff
@@ -221,6 +223,8 @@ addons/<my_module_name>/
     |-- __init__.py
     |-- <wizard_model>.py
     `-- <wizard_model>.xml
+|-- examples/
+|   |-- my_example.csv
```
dreispt: Isn't this demo data?

PR #217, opened 2016-08-25
[IMP] Exception in file naming for o2m

pedrobaeza:


moylop260:
👍

lasley:
I honestly don't like this - IMO 1:1 is much better because long files seriously suck. 

Case in point would be the existing core `sale.py` - 803 lines are being used as part of SaleOrder and SaleOrderLine, which creates an unwieldy source to work with in my opinion. I can never find where SaleOrder ends and SaleOrderLine ends.

rvalyi:
:+1: I vote for this exception. Many developers work efficiently with simple editors like Vim and switching between files constantly make loose us focus, so I vote for having some balance in splitting the files.

rvalyi:
@lasley About SaleOrder / SaleOrderLine: one thing is what Odoo SA does and unfortunately experience shows we cannot really make them improve their coding. But an other thing is what we do in the OCA. Here typically sale.py in 1000 lines in odoo/odoo and yes it sucks. But in an OCA repo your extension module would typically make a 10 line override in SaleOrder and another one in SaleOrderLine and then having both in the same file really doesn't hurt IMHO.

lasley:
@rvalyi - This exception does not consider new modules. If it exists, things like `medical.prescription.order` and `medical.prescription.order.line` in `vertical-medical` would be (and were previously) causing the same issue. IMO we have the same problem in the connectors.

rvalyi:
@lasley but then if you see some 500+ lines file you can complain during the review. What is important IMHO is not forcing the maintainers of the existing modules to spend their time doing useless splitting ceremony for no real benefit...

lasley:
@rvalyi I can't though, because the guidelines allow it. I was going to suggest some sort of addition to the exception, but we're just stacking more and more rules that everyone has to remember. 

I don't think there's a way to sway me from 👎  on this TBH, but I won't argue if I'm outvoted either. Democracy ftw

elicoidal:
I am in favor of integration even if it makes longer files because there are multiple links between the models and interactions between them.
I resent the length of the files but appreciate not to have to switch files to check obvious interactions within one object.
so :+1: with @pedrobaeza 

A word of caution though: I agree with @lasley that the set of rules is becoming longer and longer and we should be careful not to build something that becomes unmanageable, difficult to learn and by extension difficult to apply.

moylop260:
In fact, we have this rule currently in the [structure of files](https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#file-naming): `models/<main_model>.py` and `views/<main_model>.xml`
The name of the file is based on the main model, but we can have other secondary models inside.
Otherwise,  it would have been `models/<model>.py` or `views/<model>.xml` (Without `main_`)

But is better explicit than implicit.
Then I'm agree with this PR by explicit description of this case.

bealdav:
I'm ok with @lasley and @elicoidal 

> set of rules is becoming longer and longer and we should be careful not to build something that becomes unmanageable.

@rvalyi is ok when he say 

> an OCA repo your extension module would typically make a 10 line override in SaleOrder and another 10 lines override in SaleOrderLine and then having both in the same file

@lasley too

> This exception does not consider new modules. If it exists, things like medical.prescription.order

Then we could replace rules by recommendation (aka Guidelines), something around that:

> To promote readibility, developers must split python files. A good balance must be found to have code with nearest concept in the same file and avoid long files

> Avoid more X00 lines by file.

About file name, if your class are 'product.category' and 'product', then your file will be product.py, simple efficient.

Too many rules kill pleasure, enjoy and ... contributions.

my 2 cents

my 2 cents

gurneyalex:
Allowing this really kills the benefit of the one file / model rule which is to allow to easily find in which file a given model is defined / inherited.

bealdav:
I prefer Pedro update than no change at all.
:+1: for Pedro proposal.

chafique-delli:
:+1: for Pedro proposal which gives a little more flexibility to the rule.

moylop260:
We have a draw

| :+1:       | :-1: |
|------------|------|
| @rvalyi    |  @gurneyalex    |
| @elicoidal |    @lasley  |
| @moylop260 |   @bealdav  |
| @bealdav | @chafique-delli |

IMHO the draw case is polemic case and should be:
"Allow use it as you prefer"

I mean, If you have a file by model... good!
If you have a file by MAIN model... good!
Avoiding block a PR by a "polemic" case.

Similar case was with use of `.pot` files, the decision was: if you want add it or not.

elicoidal:
Agree. Let's make the decision case by case and not block PR for this 
reason if well documented/reasonable.

hparfr:
_Number of models per classes_ is part of **code quality and readability** and should be treated the same way as _number of lines of code per method_ or _choosing good names_.

The rule of thumb is common sense : there should be no exception and no strict rule.

👍  on allowing multiple models on the same file when it makes sense and the file is short

dreispt:
I am :+1:

pedrobaeza:
At the end of this discussion, we agree to relax the current rule with this exception, so I merge.

PR #218, opened 2016-08-26
Add infrastructure DNS repo

elicoidal:


tools/repos_with_ids.txt
```diff
@@ -122,3 +122,4 @@
 221|github.com/OCA/l10n-ethiopia
 222|github.com/OCA/business-requirement
 223|github.com/OCA/l10n-cambodia
+223|github.com/OCA/infrastructure-dns
```
pedrobaeza: It can't be 223, as this one is for l10n-cambodia. Have you already created the repo in runbot?

elicoidal: sorry it is 224 (stupid copy paste mistake)

pedrobaeza:
OK, we can fast-track squashing.

dreispt:
What about Transifex key?

PR #223, opened 2016-09-22
support for new style setuptools-odoo package names (setuptools-odoo 1.0.0b7)

sbidoul:
This is already running on https://wheelhouse.odoo-community.org/oca

### DIFF
--- a/tools/bdist_all_wheels.py
+++ b/tools/bdist_all_wheels.py
@@ -46,6 +46,8 @@ def branch_to_series(branch):
         return '8'
     elif branch.startswith('9.0'):
         return '9'
+    elif branch.startswith('10.0'):
+        return '10'
     else:
         raise RuntimeError("Can't determine Odoo series from %s" % branch)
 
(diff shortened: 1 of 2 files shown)


tools/bdist_all_wheels.py
```diff
@@ -34,14 +31,25 @@
...
 """
 
 
+def branch_to_series(branch):
+    if os.environ.get('SETUPTOOLS_ODOO_LEGACY_MODE'):
+        return ''
+    if branch.startswith('8.0'):
+        return '8'
+    elif branch.startswith('9.0'):
+        return '9'
```
pedrobaeza: We should include here 10 and 11 as well as we have clear that those 2 will be in the future (or not? :stuck_out_tongue_winking_eye: )

sbidoul: For sure it will be in the future. There will be another PR for that. I'm wating on https://github.com/odoo/odoo/pull/13237 for 10.0.

pedrobaeza:
Please check Travis

sbidoul:
Merging and installing on OCA server

PR #224, opened 2016-09-23
Prevent ImportError by default.

Yajo:
### Fixes / new features

Tired of fixing `ImportError`s [again](https://github.com/OCA/account-financial-reporting/pull/229) and [again](https://github.com/OCA/l10n-spain/pull/255) and [again](https://github.com/OCA/reporting-engine/pull/64) and [again](https://github.com/OCA/reporting-engine/pull/62)...

@Tecnativa

### DIFF
--- a/template/module/__init__.py
+++ b/template/module/__init__.py
@@ -1,9 +1,14 @@
 # -*- coding: utf-8 -*-
 # Copyright <YEAR(S)> <AUTHOR(S)>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
-
-from . import models
-from . import wizards
-from . import controllers
-from . import report
-from .hooks import pre_init_hook, post_load, post_init_hook, uninstall_hook
+try:
+    from . import models
+    from . import wizards
+    from . import controllers
+    from . import report
+    from .hooks import pre_init_hook, post_load, post_init_hook, uninstall_hook
+except ImportError:
+    import logging
+    _logger = logging.getLogger(__name__)
+    _logger.info("ImportError raised while loading module.")
+    _logger.debug("ImportError details:", exc_info=True)


sbidoul:
I'm not favorable to hide real errors. :-1:

Yajo:
What am I hiding?

sbidoul:
You are transforming potential real bugs (eg a mistyped import statement) into info/debug level logs. 

If you want to solve spurious CI errors due to missing dependencies, let's work on the root cause instead: https://github.com/OCA/maintainer-quality-tools/pull/343 (BTW, I'm willing to work on this during the october sprint)

danimaribeiro:
What's the purpose of loading a module in a inconsistent state?
:-1:

lmignon:
:-1:

pedrobaeza:
You're mixing things: you're not loading modules (they're not even installed), it's Odoo who loads them, at least the Python part. With this, you prevent your server stalls at loading. And it happens when you add a repository to your addons_path for using one of the modules, but you don't want to add the corresponding dependencies for non-installed modules.

Second: I see good all the pip promotion, but there's still a lot of people who don't want or can't use this method, so we shouldn't make this fail in that cases.

pedrobaeza:
Said that, we can lower the try at a deep level for not being so "aggressive" and this masks other programming errors.

moylop260:
@Yajo 
I want understand your point first.

Do you have a real case showing a `ImportError` with `from . import something`?

moylop260:
I understand your point (I saw the url of `again`)

FYI we have a guideline for external dependencies
https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#external-dependencies

You could use it to add this case for template module.

moylop260:
Maybe we can create a new pylint check to validate this case.
Whitelist of allowed imports (internal libraries).
If import not in whitelist then check if there is a try-except-ImportError-logger

Yajo:
I understand the downvotes, but certainly we need a solution for https://github.com/OCA/maintainer-tools/pull/224#issue-178840620. @moylop260's suggestion would be better indeed, however I'm not sure on how to do that...

About pip... I'm not sure on how the pip installing works, but I feel it would harden the task of merging PRs on your production environments, which is something we basically cannot live without. @sbidoul Is there a way to combine both of those needs?

sbidoul:
@Yajo 

>  I feel it would harden the task of merging PRs on your production environments

Merging PR in real world deployment is absolutely necessary indeed. Pip does not make it harder, on the contrary:

- pip has a develop/editable mode to install unreleased code (`pip install -e repo/setup/addon_name`)
- there is [git-aggregator](https://github.com/acsone/git-aggregator) to manage merges in a repeatable manner

It's true it's a change of habitude and it require some learning, but we are very happy we did this change at acsone after several years using buildout. Do good thing is that what you learn with pip is applicable almost everywhere python is used.

moylop260:
FYI I have created the check for pylint-odoo: https://github.com/OCA/pylint-odoo/pull/68

Yajo:
Thanks to both! Closing in favour of the pylint alternative

PR #227, opened 2016-10-04
Changes in order to push 10.0 branches to OCA projects

StefanRijnhart:
[RFR] Check \_\_manifest\_\_.py when generating addons tables
and migrating branches
[IMP] Rename manifest to \_\_manifest\_\_.py for 10.0 branches
[FIX] Fetch the tree of the target branch, not of the project main branch
[IMP] Fix missing comma when adding dictionary entry
[IMP] Sort modules by name on the migration issue
[ADD] Delete setup addon directories on branch migration

### DIFF
--- a/tools/gen_addons_table.py
+++ b/tools/gen_addons_table.py
@@ -23,6 +23,7 @@ import re
 
 
 MARKERS = r'(\[//\]: # \(addons\))|(\[//\]: # \(end addons\))'
+MANIFESTS = ('__openerp__.py', '__manifest__.py')
 
 
 class UserError(Exception):
@@ -93,8 +94,12 @@ def gen_addons_table():
     rows_available = []
     rows_unported = []
     for addon_path, unported in addon_paths:
-        manifest_path = os.path.join(addon_path, '__openerp__.py')
-        if os.path.isfile(manifest_path):
+        for manifest_file in MANIFESTS:
+            manifest_path = os.path.join(addon_path, manifest_file)
+            has_manifest = os.path.isfile(manifest_path)
+            if has_manifest:
+                break
+        if has_manifest:
             manifest = ast.literal_eval(open(manifest_path).read())
             addon_name = os.path.basename(addon_path)
             link = '[%s](%s/)' % (addon_name, addon_path)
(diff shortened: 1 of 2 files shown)


StefanRijnhart:
cc @pedrobaeza, @sbidoul

StefanRijnhart:
See https://github.com/StefanRijnhart/server-tools/commits/10.0 for an example of the commits that this version of the script creates.

dreispt:
Looks good :+1:
Do  we want to also delete the unported modules?

pedrobaeza:
I miss the sorting of the module list in the issue, but I can add it later.

sbidoul:
Looks good :+1:

StefanRijnhart:
@pedrobaeza Hold on, let me do the sorting.

StefanRijnhart:
@dreispt we always keep the unported modules so that the changes of the code migration can be reviewed relatively transparently, right?

StefanRijnhart:
@pedrobaeza https://github.com/OCA/maintainer-tools/pull/227/commits/87d79bfa1b758de0e9603e2271b395668166c868

Leaving this to work in progress because I want to do one final test.

StefanRijnhart:
So the good news is that we don't have to rename the manifest file, because the server does not attempt to load modules set to uninstallable. The bad news is that not all modules are set to uninstallable. To be continued.

StefanRijnhart:
Current version is good to go. See the updated PR description above. Also updated is https://github.com/StefanRijnhart/server-tools/commits/10.0.

StefanRijnhart:
Oh wait, maybe we want to rename the manifests to __manifest__.py?

pedrobaeza:
We can let that task to be done in the migration process.

sbidoul:
We need to remove the setup directories for imported add-ons but I can do it after running this script.

StefanRijnhart:
I'll have a look, I am now in shape for working with the Github API ;-)

StefanRijnhart:
Now removing setup addon dirs (@sbidoul, I hope this is alright) and renaming \_\_openerp\_\_.py to \_\_manifest\_\_.py.

sbidoul:
@StefanRijnhart removing setup addon dirs is fine. Can't comment on the how as I'm not so much into the github api.

pedrobaeza:
I'm afraid you can't remove a full folder directly. You have to remove file per file with the API.

PR #228, opened 2016-10-04
Allow to split main.py in severeal files

antespi:
@simahawk and me purpose to allow spliting controllers/main.py in several files when there are many controller methods or when big ones. And this case can be often in website modules.

We've discussed this in OCA Sprint in Brussels. Now this is our proposal and maybe @lasley, @pedrobaeza, @dreispt and @gurneyalex have something to say about this.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -132,7 +132,8 @@ For model named `<main_model>` the following files may be created:
 * `templates/<main_model>.xml`
 * `views/<main_model>.xml`
 
-For `controller`, the only file should be named `main.py`.
+For `controller`, if there is only file should be named `main.py`. If there
+are several controller methods you can split them in several files.
 
 For `static files`, the name pattern is `<module_name>.ext` (i.e.
 `static/js/im_chat.js`, `static/css/im_chat.css`, `static/xml/im_chat.xml`,
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -132,8 +132,9 @@ For model named `<main_model>` the following files may be created:
 * `templates/<main_model>.xml`
 * `views/<main_model>.xml`
 
-For `controller`, if there is only file should be named `main.py`. If there
-are several controller methods you can split them in several files.
+For `controller`, if there is only one file it should be named `main.py`.
+If there are several controller classes or functions you can split them into
+several files.
 
 For `static files`, the name pattern is `<module_name>.ext` (i.e.
 `static/js/im_chat.js`, `static/css/im_chat.css`, `static/xml/im_chat.xml`,


lasley:
Let me preface this with the fact that I hate large source files, and was firmly against the combination of models in the same file despite them being related.

The problem I see here is that we do not have a controller registry as we do with models. This forces us to `from openerp.addons.my_addon.controlles.main import Whatever`, meaning we have to know the entire import path including file and controller name.

Due to the way that Odoo packaging works, I don't get autocomplete on these imports until I have traversed into the file (using Komodo IDE). Currently, I can guess the import path because it is always the same. If we are to change the single file rule for controllers, I have to look at the source of the parent module to determine what the import path actually is.

Ever since we spoke about this earlier, I have been pondering how we would make a controller registry and have determined that's a mute point because we would need to add some sort of `_name` attribute to all controllers in existence in order to accommodate - which is definitely out of the question.

My other idea is that we can allow the splitting of files in instances where it makes sense, but still provide a `main.py` that functions similar to the init file, aggregating the classes into the main so that we can still import in the same manor. This seems kind of redundant though, so I'm not so sure that it's a great solution either.

Anyways, theres my two cents. Anyone else have some thoughts on this, or possible solutions to allow my laziness while appeasing the need for small source files?

simahawk:
IMHO if you are extending/inheriting/whatever you still need look at the code to know what to import right? 
To do that you check `main.py`. At that point **you know where you are** and from which file you import from. Plus, if you are looking into `controllers/` folder and you see a file called `form.py` than you save time, because you know that you find form-related code into it.

CONTRIBUTING.md
```diff
@@ -132,7 +132,8 @@ For model named `<main_model>` the following files may be created:
 * `templates/<main_model>.xml`
 * `views/<main_model>.xml`
 
-For `controller`, the only file should be named `main.py`.
+For `controller`, if there is only file should be named `main.py`. If there
```
simahawk: "if there is only file should be named `main.py`" should be "if there is only one file it should be named `main.py`"

CONTRIBUTING.md
```diff
@@ -132,7 +132,8 @@ For model named `<main_model>` the following files may be created:
...
+are several controller methods you can split them in several files.
```
simahawk: "If there are several controller methods you can split them in several files." should be "If there are several controller classes or functions you can split them into several files."

antespi: Done

pedrobaeza:
I'm also in favor of making the split in several files following the same principle as in models, views and so on, and this redaction allows to be an option, not mandatory, so everyone wins :smile:

simahawk:
shall we merge this? :)

leemannd:
:+1: As said by @antespi it becomes really usefull in websites.

lasley:
😢

PR #230, opened 2016-10-12
external_dependency in tests files must not be protected

lmignon:


moylop260:
👍

hbrunn:
in that case, we should also adapt the pylint script not to test for this. web 8.0 is currently red because if this

pedrobaeza:
I have seen the corresponding PR, so I merge this one and review the PR to pylint-odoo.

PR #241, opened 2016-11-18
[IMP] template: Include funders

pedrobaeza:
To include in the template the possibility of adding funders of the module


antespi:
:+1:

PR #242, opened 2016-11-22
[ADD] template: Abstract model test strategy

lasley:
* Add a sample of a testing strategy for an abstract model

We may want to rename some things, or possibly change the structure, but this is a good one to include IMO. 

This is pulled from @obulkin resolution of OCA/web#490 as inspired by @hbrunn

### DIFF
--- a/template/module/tests/test_abstract_something.py
+++ b/template/module/tests/test_abstract_something.py
@@ -1,0 +1,53 @@
+# -*- coding: utf-8 -*-
+# Copyright <YEAR(S)> <AUTHOR(S)>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from openerp import api, models
+from openerp.tests.common import SingleTransactionCase
+
+
+class AbstractSomethingTester(models.Model):
+    """ It provides a real model object to test the abstract with """
+    _name = 'abstract.something.tester'
+    _description = 'Abstract Something Tester'
+    _inherit = 'abstract.something'
+
+
+class TestAbstractSomething(SingleTransactionCase):
+    @classmethod
+    def _init_test_model(cls, model_cls):
+        """ It builds a model from model_cls in order to test abstract models
+
+        Args:
+            model_cls: (openerp.models.BaseModel) Class of model to initialize
+        Returns:
+            Model instance
+        """
+        model_cls._build_model(cls.registry, cls.cr)
+        model = cls.env[model_cls._name].with_context(todo=[])
+        model._prepare_setup()
+        model._setup_base(partial=False)
+        model._setup_fields(partial=False)
+        model._setup_complete()
+        model._auto_init()
+        model.init()
+        model._auto_end()
+        return model
+
+    @classmethod
+    def setUpClass(cls):
+        super(TestAbstractSomething, cls).setUpClass()
+        cls.registry.enter_test_mode()
+        cls.old_cursor = cls.cr
+        cls.cr = cls.registry.cursor()
+        cls.env = api.Environment(cls.cr, cls.uid, {})
+        cls.test_model = cls._init_test_model(AbstractSomethingTester)
+
+    @classmethod
+    def tearDownClass(cls):
+        cls.registry.leave_test_mode()
+        cls.registry[cls.test_model._name]._abstract = True
+        cls.registry[cls.test_model._name]._auto = False
+        cls.cr = cls.old_cursor
+        cls.env = api.Environment(cls.cr, cls.uid, {})
+        super(TestAbstractSomething, cls).tearDownClass()
--- a/template/module/models/abstract_something.py
+++ b/template/module/models/abstract_something.py
@@ -1,0 +1,10 @@
+# -*- coding: utf-8 -*-
+# Copyright <YEAR(S)> <AUTHOR(S)>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from openerp import models
+
+
+class AbstractSomething(models.AbstractModel):
+    _name = "abstract.something"
+    _description = "Abstract Somethings"
--- a/template/module/tests/__init__.py
+++ b/template/module/tests/__init__.py
@@ -3,3 +3,4 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from . import test_something
+from . import test_abstract_something


lasley:
Hmmm or maybe we just include ourselves a helper in the new `oca` lib we're planning in OCA/server-tools#615

lmignon:
@lasley In some cases I create test models on demand. Here it's the code used to only create these models when I need them in my tests...
The method https://github.com/acsone/alfodoo/blob/9.0/cmis_field/tests/common.py#L10 (V9)
https://github.com/acsone/alfodoo/blob/10.0-dev/cmis_field/tests/common.py (V10)

And a call to the method https://github.com/acsone/alfodoo/blob/9.0/cmis_field/tests/common.py#L29

lasley:
Oh @lmignon you're taking it a step further than I am here by building out the fields and all that too. I like! Will add 😄

lasley:
Code has been refactored in light of @lmignon suggestions, and I moved the `ir.model` creation into its own method in order to make it clear that it is not necessary in all cases.

template/module/tests/test_abstract_something.py
```diff
@@ -0,0 +1,64 @@
...
+        """
+        registry = cls.env.registry
+        cr = cls.env.cr
+        inst = model_cls._build_model(registry, cr)
+        model = cls.env[model_cls._name].with_context(todo=[])
+        model._prepare_setup()
+        model._setup_base(partial=False)
+        model._setup_fields(partial=False)
+        model._setup_complete()
+        model._auto_init()
```
obulkin: This is the method that actually creates a table for the model, and it will break the isolation provided by `TransactionCase`: https://github.com/odoo/odoo/blob/18456b9d42769a522adf86183746b49735e9d8b3/odoo/models.py#L2535. Unfortunately, I don't see a way to proceed beyond the `_setup_complete()` step without introducing this issue, which defeats the entire purpose of building a model in this way.

lasley: Hmmm yep good call, thanks @obulkin. I can't think of an instance where the table be required outside of the test transaction, so I believe you're correct in being able to remove it. Well, that and the whole necessity thing.

lmignon: Yep I've forgotten to mock cr.commit...

lasley: Ohhhh mock `cr.commit`! And there's the solution, I like this better than removing the table creation. Thanks @lmignon

Yajo: You can [use a test cursor like the one used in HttpCase instead](https://github.com/odoo/odoo/blob/10.0/odoo/tests/common.py#L246). [It mocks cr.commit by itself](https://github.com/odoo/odoo/blob/fc2e80cb4bcc450762c7ac5cb82a3e2d88062b38/odoo/sql_db.py#L445-L447).

lasley: TIL, thanks @Yajo. So glad I made this PR, we're really drilling to the true solution!

obulkin: Ooh. Nice find, @Yajo. I wonder why `TransactionCase` doesn't just use the test cursor/savepoint approach.

Yajo: [It wasn't so nice at the beginning](https://github.com/odoo/odoo/issues/12237), it's just [getting nicer now](https://github.com/odoo/odoo/issues/13902). 😆 

FTR, [`TransactionCase` makes a rollback instead](https://github.com/odoo/odoo/blob/10.0/odoo/tests/common.py#L155). Not sure why, but that's what it does.

template/module/tests/test_abstract_something.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+class AbstractSomethingTester(models.Model):
+    """ It provides a real model object to test the abstract with """
+    _name = 'abstract.something.tester'
+    _description = 'Abstract Something Tester'
+    _inherit = 'abstract.something'
+
+
+class TestAbstractSomething(TransactionCase):
```
lasley: This has to be `SingleTransactionCase` or `SavepointCase` to work as is. Otherwise, the generation methods need to be moved to `setUp`. Maybe make two example classes instead.

template/module/tests/test_abstract_something.py
```diff
@@ -0,0 +1,64 @@
...
+        return inst
+
+    @classmethod
+    def setUpClass(cls):
+        super(TestAbstractSomething).setUpClass()
+        # mock commit since it"s called in the _auto_init method
+        cls.cr.commit = mock.MagicMock()
+        cls._init_test_model(AbstractSomethingTester)
+
+    def _create_ir_model(self, model_cls):
```
lasley: This isn't needed. Somewhere in all the new stuff, this is already done

template/module/tests/test_abstract_something.py
```diff
@@ -0,0 +1,64 @@
...
+# Copyright <YEAR(S)> <AUTHOR(S)>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+import mock
+
+from openerp import models
+from openerp.tests.common import TransactionCase
+
+
+class AbstractSomethingTester(models.Model):
```
lasley: This needs to be a TransientModel in order to not trigger security warning

template/module/tests/test_abstract_something.py
```diff
@@ -0,0 +1,64 @@
...
+
+    @classmethod
+    def _init_test_model(cls, model_cls):
+        """ It builds a model from model_cls in order to test abstract models
+
+        Note that this does not actually create a table in the database, so
+        there may be some unidentified edge cases.
+
+        Args:
+            model_cls: (openerp.models.BaseModel) Class of model to initialize
```
lasley: Note from https://github.com/OCA/server-tools/pull/681/files/aebfc5578c5e8d4fdca5065465d6066d588904ba#r98303294

There are some issues with this docblock:

- As currently written, the method does actually create a table for the model. That happens as part of `_auto_init()`
- The var name in the `Returns` section is wrong

obulkin:
I've updated this PR to incorporate the suggestions in the comments, as well as a recent realization about the correct way to use the registry test mode. The current approach uses ``SingleTransactionCase`` but can be modified to use ``TransactionCase`` where needed by moving all the logic into the appropriate instance methods (e.g. ``setUp`` instead of ``setUpClass``). It does not work with ``SavepointCase`` due to some sort of conflict between the savepoints introduced by this class and the ones used by the registry test mode in place of commits.

template/module/tests/test_abstract_something.py
```diff
@@ -0,0 +1,53 @@
...
+        cls.cr = cls.registry.cursor()
+        cls.env = api.Environment(cls.cr, cls.uid, {})
+        cls._init_test_model(AbstractSomethingTester)
+
+    @classmethod
+    def tearDownClass(cls):
+        cls.env.registry.leave_test_mode()
+        cls.cr = cls.old_cursor
+        cls.env = api.Environment(cls.cr, cls.uid, {})
+        super(TestAbstractSomething, cls).tearDownClass()
```
Yajo: Are you allowed to call `super()` on `classmethod`s? 😕

lasley: A damned fine question - I just had to [look this up](https://docs.python.org/2/library/functions.html#super) to find out. Super accepts an `object-or-type` as the second argument for binding. It even specifically mentions classmethods:

> If the second argument is a type, issubclass(type2, type) must be true (this is useful for classmethods).

obulkin: I figured this was fine because it was working in a module that I just updated. 😆  Good to see official confirmation from the docs though.

obulkin:
I've discovered another issue with this approach, which comes up when the module with the abstract model test has a child module that is installed and tested at the same time. The tester model remains in the registry after testing is finished and triggers an error during the install process for the child module. The latest commit addresses this by manually removing it from the registry during test cleanup.

lasley:
Thanks for the enlightenment Oleg. Interesting how many edge cases there are with this

obulkin:
I've updated this PR again to address two more edge cases:

- If the tester model is simply removed from the registry and a subsequent module install causes one of its parent models to be modified, Odoo will attempt to rebuild the tester model and will throw an error when it can't be found. Instead, it can be made abstract, which allows this process to go through while preventing errors caused by the absence of an associated DB table.
- The tester model should inherit from ``models.Model`` instead of ``models.TransientModel`` since its lack of a DB table will cause an error when transient model data is cleaned up.

lasley:
Thanks @obulkin 

Does anyone want to review this one? It'd be nice to have an official abstract test strategy, even if we're still learning on it. Otherwise this PR is >1 year, so I move for a close due to lack of interest - regardless of use.

Yajo:
I approved it already

lasley:
At this point it's Oleg's code. Does my 👍  count for this?

Yajo:
I'd say yes 😄. Please squash a bit before merge 😊

lasley:
@obulkin - please squash all the commits and author it to yourself. It was your code that I based on anyways.

obulkin:
@lasley - Sure thing. I've squashed down to one commit and changed the author.

PR #243, opened 2016-11-24
[REF] tx_pull: Remove "date_range" filter

moylop260:
I think that this piece of code is for faster debug but not for production case.

Is right @pedrobaeza?

Could you share us the production code?

### DIFF
--- a/tools/tx_pull.py
+++ b/tools/tx_pull.py
@@ -224,8 +224,6 @@ class TransifexPuller(object):
         resources = tx_project_api.resources().get()
         for resource in resources:
             print "Checking resource %s..." % resource['name']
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Uhm, that's something I don't remember. Who introduce that? Maybe it was introduce to avoid an error on that concrete module?

moylop260:
> Maybe it was introduce to avoid an error on that concrete module?

This theory is valid for `if resource['slug'] == 'date_range':` but for this case is `if resource['slug'] != 'date_range':`
I think that It was a faster test that was merged.

pedrobaeza:
Yeah, you're right. I'll fast-track this then.

PR #244, opened 2016-11-29
[IMP] tx_pull: Handle rate limits

pedrobaeza:
This improves the wait times for avoiding rate limits using webservices until limit error codes are obtained. 

I have also protected all the possible failures in the process, iterating over a full synchronization on all the OCA projects. With this, there are enough guarantees that a full sync process finishes.

### DIFF
--- a/tools/tx_pull.py
+++ b/tools/tx_pull.py
@@ -246,8 +246,6 @@ class TransifexPuller(object):
                         self.tx_api.project(project_slug).get
                     )
                     tx_projects.append(tx_project)
-                except (KeyboardInterrupt, SystemExit):
-                    raise
                 except exceptions.HttpNotFoundError:
                     print "ERROR: Transifex project slug '%s' is invalid" % (
                         project_slug
@@ -297,8 +295,6 @@ class TransifexPuller(object):
                         tx_lang = wrap_tx_call(
                             tx_resource_api.translation(lang).get,
                         )
-                    except (KeyboardInterrupt, SystemExit):
-                        raise
                     except (exceptions.HttpClientError,
                             exceptions.HttpServerError):
                         tx_lang = False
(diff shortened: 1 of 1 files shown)


tools/tx_pull.py
```diff
@@ -191,20 +242,24 @@ def process_projects(self, projects=None):
             # Check that provided projects are correct
             for project_slug in projects:
                 try:
-                    tx_project = self.tx_api.project(project_slug).get()
+                    tx_project = wrap_tx_call(
+                        self.tx_api.project(project_slug).get
+                    )
                     tx_projects.append(tx_project)
                 except (KeyboardInterrupt, SystemExit):
```
moylop260: Duplicated exceptions from `wrap_tx_call` method.

tools/tx_pull.py
```diff
@@ -107,13 +108,62 @@ def get_parser():
...
+              'associated to the GitHub account is not public'))
     parser.add_argument(
         '-t', '--target-org', dest='target',
-        help=('By default, translation are committed in github on OCA. This '
+        help=('By default, translation are committed in GitHub on OCA. This '
               'arg lets you provide an alternative org'))
     return parser
 
+
+def wrap_tx_call(func, args=None, kwargs=None):
```
moylop260: What about use `*args, **kwargs`?
I mean
screen shot 2016-11-29 at 2 42 57 pm

pedrobaeza: This method is calling the other passed as first argument, so it's not exactly the same. I think this one is good to go this way as arguments are not mixed and clearly passed.

tools/tx_pull.py
pedrobaeza: Removed

tools/tx_pull.py
moylop260: What is wrong with the first argument?

screen shot 2016-11-29 at 2 52 34 pm

pedrobaeza: OK, I see. Yeah, it's also possible that way, but if you don't mind, I have invested a lot of time here, so I prefer to keep it this way as this is only a cosmetic change

moylop260: Ok I got it

With this change you could avoid invoke with:
`mym(func, args=(), kwags={})` and use directly `mym(func, arg1, arg2, kwar1='value')`

But I understand your point.
Maybe other PR to change it
Thanks!

tools/tx_pull.py
```diff
@@ -107,13 +108,62 @@ def get_parser():
...
+    """Intercept all TX calls to wait when the API rate limit is reached."""
+    while True:
```
lasley: I'm not entirely sure architecture of how this is run in production, but is this threaded at all? This could cause an infinite loop if Transifex goes down I think, which could cause some stacking of sockets

pedrobaeza: This is a shell script, so no threading here. The exit condition is on the raises or the normal return exit, so no worry about being stacked.

gurneyalex:
nice ! Thanks a lot.

PR #250, opened 2016-12-26
[REF] CONTRIBUTING: Remove commit for with registry

moylop260:
Based on @Yajo comment: https://github.com/OCA/web/pull/506#pullrequestreview-14301433

And odoo comments and code: https://github.com/odoo/odoo/blob/fc2e80cb4bc/odoo/sql_db.py#L384-L391

For merger: Squash and merge

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -690,8 +690,7 @@ Unless:
             # with_env replace original env for this method
             # A good comment here of why this isolated transaction is required.
             self.with_env(new_env).write({'name': 'hello'})  # isolated transaction to commit
-        # You don't need close your cr because is closed when finish "with"
-        # You don't need commit your cr because is commited when finish "with"
+        # You don't need to close nor to commit your cursor as they are done when exiting "with" block
     # You don't need clear caches because is cleared when finish "with"
   ```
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -688,9 +688,10 @@ Unless:
             # Create a new environment with new cursor database
             new_env = api.Environment(new_cr, self.env.uid, self.env.context)
             # with_env replace original env for this method
+            # A good comment here of because is required this transaction isolated.
```
pedrobaeza: s/because/why

pedrobaeza: Full sentence should be "A good comment here of why this isolated transaction is required."

CONTRIBUTING.md
```diff
@@ -688,9 +688,10 @@ Unless:
             # Create a new environment with new cursor database
             new_env = api.Environment(new_cr, self.env.uid, self.env.context)
             # with_env replace original env for this method
+            # A good comment here of why this isolated transaction is required.
             self.with_env(new_env).write({'name': 'hello'})  # isolated transaction to commit
-            new_env.cr.commit()  # Don't show a invalid-commit in this case
         # You don't need close your cr because is closed when finish "with"
```
pedrobaeza: Instead of putting 2 lines, rephrase this one as:
```
# You don't need to close nor to commit your cursor as they are done when exiting "with" block
```

moylop260: Fixed thanks @pedrobaeza

PR #251, opened 2016-12-30
[IMP] Bug tracker wording in ReadMe

lasley:
Minor fix of the wording as proposed in https://github.com/OCA/server-tools/pull/642#discussion_r94187646 - been meaning to submit this for a while, but the comment was the final straw 😛


moylop260:
Fast tracking because is a minor and typo fix

PR #253, opened 2017-01-07
[IMP] Upgrade datafile API

lasley:
This upgrades the view and data files to use the new `<odoo>` syntax instead of `<openerp><data>`

I also added a note about `noupdate="1"` & Changed the indent level. 

The indentation level may be undesired, but I also find it weird that we're not indenting properly. I'll change it back if requested, just figured I would propose that while I was already editing the files.

Per discussion in https://github.com/LasLabs/vertical-medical/pull/81#discussion_r95057329

cc @jcdrubay

### DIFF
--- a/template/module/data/model_name_data.xml
+++ b/template/module/data/model_name_data.xml
@@ -2,12 +2,11 @@
 <!-- Copyright <YEAR(S)> <AUTHOR(S)>
      License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
 
-<openerp>
-<data>
+<!-- Note that ``<odoo noupdate="1">`` will also work here. -->
+<odoo>
 
-<record id="my_data_record" model="model.name">
-    <!-- My record data -->
-</record>
+    <record id="my_data_record" model="model.name">
+        <!-- My record data -->
+    </record>
 
-</data>
-</openerp>
+</odoo>
--- a/template/module/security/some_model_security.xml
+++ b/template/module/security/some_model_security.xml
@@ -2,30 +2,29 @@
 <!-- Copyright <YEAR(S)> <AUTHOR(S)>
      License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
 
-<openerp>
-<data>
+<!-- Note that ``<odoo noupdate="1">`` will also work here. -->
+<odoo>
 
-<record id="category" model="ir.module.category">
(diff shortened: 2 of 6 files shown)


pedrobaeza:
The problem again here is that this template is for all versions, and version 8 and 7 (still active) won't work with this template.

lasley:
v7 is EOL, so I don't think we should still consider it in the templates. 

Maybe we should somehow represent both in the template then? We're telling contributors to make changes, but our documentation/templates are not providing this example due to the evolving nature of the API. I think this problem is deeper than just the data tags TBH

pedrobaeza:
I agree that we should reflect in documentation. Possibilities:

* Put in comments the deprecated style inside the same file.
* Have several files with a suffix like `..._v9_and_above.xml` (or the contrary) to reflect changes depending on version.
* I suggested some time ago to have several branches for each version with specific templates.

lasley:
I was on board with your original multi-branch suggestion, but I think I remember us being pretty heavily outnumbered there.

I like the different files then, although something maybe a bit more concise? If there is no _v* on the files, one can assume it works in all Odoo versions. If there's versions, one can assume that the file will be compatible with Odoo versions matching or exceeding that version - up until the next versioned file. E.g.:

* Given directory with files: `model_name_v7.xml` & `model_name_v9.xml`
* V9 would use v9.xml
* V10 would use v9.xml
* V8 would use v7.xml

pedrobaeza:
With that formula for sure there will be opened issues claiming where is the `model_name_v8.xml`, and duplicating files identical are not the best option. That's why I told about `model_name_v9_and_above.xml`, but I make an alternative precision:

* For already known intervals, we put it explicitly: `model_name_v5-v8.xml`.
* For the current implementation, we put it with nothing: `model_name.xml`.

This way, people will look for sure for their explicit file if the version matches. If not, people will go to the agnostic name one. This way, we prevent to rename the files on each version.

jcdrubay:
I also like the "multi-files solutions" the best.

I imagine that the principle "explicit better than implicit" should apply, so 3 options:
 - `model_name_v9_and_below.xml`
 - `model_name_v5_to_v8.xml`
 - `model_name_v9_and_above.xml`

lasley:
We're good to move forward with this one if I rebase, correct?

pedrobaeza:
Closing this as outdated.

PR #260, opened 2017-02-24
remove copyright from __init__.py

leemannd:
### Milestone (Odoo version)
- All

### Module(s)
- All

### Fixes / new features
- Do we want to keep copyright into '__init__.py' files? Depending on reviewers, you can be asked to remove/put copyright from 'init' files. A good point that goes in this direction is that it doesn't contains code.

### DIFF
--- a/template/module/__init__.py
+++ b/template/module/__init__.py
@@ -1,5 +1,4 @@
 # -*- coding: utf-8 -*-
-# Copyright <YEAR(S)> <AUTHOR(S)>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from . import models
--- a/template/module/controllers/__init__.py
+++ b/template/module/controllers/__init__.py
@@ -1,5 +1,4 @@
 # -*- coding: utf-8 -*-
-# Copyright <YEAR(S)> <AUTHOR(S)>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from . import main
--- a/template/module/models/__init__.py
+++ b/template/module/models/__init__.py
@@ -1,5 +1,4 @@
 # -*- coding: utf-8 -*-
-# Copyright <YEAR(S)> <AUTHOR(S)>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from . import model_name
--- a/template/module/report/__init__.py
+++ b/template/module/report/__init__.py
@@ -1,5 +1,4 @@
 # -*- coding: utf-8 -*-
-# Copyright <YEAR(S)> <AUTHOR(S)>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from . import report_name
--- a/template/module/tests/__init__.py
+++ b/template/module/tests/__init__.py
@@ -1,5 +1,4 @@
 # -*- coding: utf-8 -*-
-# Copyright <YEAR(S)> <AUTHOR(S)>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from . import test_something
--- a/template/module/wizards/__init__.py
+++ b/template/module/wizards/__init__.py
@@ -1,5 +1,4 @@
 # -*- coding: utf-8 -*-
-# Copyright <YEAR(S)> <AUTHOR(S)>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from . import wizard_model


leemannd:
@pedrobaeza We could also remove the license from these files?

elicoidal:
Out of curiosity, do we need to keep the other 2 lines?

I am in favor for this change but it creates an exception in the rule of "short headers everywhere" which we need to explain to contributors

pedrobaeza:
@leemannd it's said in FSF that license should be in all source files, so I'm not sure if we can remove it, although they are extra bits that doesn't worth.

@elicoidal, I don't get you. Shorter header than this can't be :wink:

elicoidal:
@pedrobaeza I mean, we would have now a rule with exception:
* Before: "everywhere short headers"
* Now: "Everywhere short headers except in init.py where you have another kind of header"

pedrobaeza:
Well, we can talk about a short header without copyrights...

leemannd:
@pedrobaeza I just looked at the FSF and what I understood the the license should be putted in all files, but also the copyright.
@elicoidal It is already something asked by some reviewers.

pedrobaeza:
Uhm, then interpreting this firmly, it's correct as is. I don't know what to do then. I prefer to remove it obviously. Maybe @yvaucher can say, as he was the one who interpreted in the past about the "Copyright" word (if I don't remember bad).

yvaucher:
As it was said the recommendation is a "should" be in all source files.

What some of us do was to remove completely the header from `__init__.py` files where there is no imports.

The main reasons are:

* Don't overload a 1 line file.
* There is no logic code in there
* No questionable addition of copyright in import file
* People who will read the code of a module might read import files but if they does it gives them no information they will have to read the manifest and or the other `.py` files, thus they will see copyrights and licenses sooner or later.
* Smaller diffs
* Make things simple for `__init__.py` files

The only cons I see are:

* not being strict toward a "should"
* `__init__.py` are treated differently than other python files.

lasley:
The reasoning behind the recommendation of both license and copyrights being included in all source files is in the event that the files are distributed outside of their normal packaging (in this case, the module).

I personally do not think that we should override this, even for init files. The argument could be made that the init file is worthless without the files it is representing, which is perfectly valid from a real-world standpoint. My main issue is with adding exceptions for things so small.

I would instead prefer that we add a rule about Copyrights in general. If you don't edit a file, you can't claim copyright on it. Therefore, if we had a rule to explicitly disallow the addition of copyrights in files that have not been otherwise edited, we would implicitly stop the expansion of the init file copyrights.

My personal feeling is that I would prefer to not have them, because cleaner is better and less is more. I feel like making it a hardline rule though just adds complexity.

That's my two cents anyways!

dreispt:
I agree with @lasley , except for one thing:
Editing a file does not grant copyright.
Copyright is due from original work.
For example, simple module migrations do not grant copyright to the contributor.
But adding a significant feature to an existing module does grant copyright.
I concur that we need some statement explaining copyright to contributors, and how we should deal with it in OCA code.

lasley:
@dreispt - It was my understanding that the copyright is held line by line, and the headers were a representation of those lines.

I like this idea of the copyright header being original author or significant feature allowing for a copyright header addition. Maybe this should also be the criteria for addition into the `author` key in the manifest too? Kill two birds with one stone?

dreispt:
@lasley Yes, I believe so.
Take an example regarding books: when I write a book I'm the sole author, even if there is an editor rewriting a lot of my text.
And if the book is translated to another language, I'm still the only author, despite all the translation work and the full rewrite.
AFAICT authorship requires a "creative work".

dreispt:
Relevant reading: https://en.m.wikipedia.org/wiki/Derivative_work

bealdav:
:+1: to remove copyright in \_\_init\_\_ template

pedrobaeza:
OK, I think we all agree to remove copyright on these files, isn't it?

The other question about authorship, I don't agree on @dreispt's interpretation. You can "only" migrate a module, but this may require a heavy work of adaptation, investigating core's source code for the needed changes (maybe not too much, but at least it's needed to see), perhaps test creation... This is very simple to see if the migration is only to change what migration guide says (s/openerp/odoo and so on), or it has required a bit more, so the co-authorship should be discussed for each case.

dreispt:
The correct thing is to leave the copyright notice in init files.
But I agree that that can also seem pointless when these files contain no actual logic, so it's acceptable to have them without copyright.

Having said that, the OCA guidelines can't go against "the correct thing", so I'm :-1: to include that removal an an OCA guideline.

leemannd:
Fair enough

pedrobaeza:
@dreispt the problem is which copyright you put? All the contributors? The first? I don't think it's not correct to remove it from that files, as they don't contain any logic to be copyrighted.

dreispt:
In the case of addons init file, its true they are just "helpers" and contain no real "original work" as you say.
Copyright belongs to the authors of the code.

Whatever way we go, it's not really that important, and I personally don't care if the copyright lines are not there.
But since the GPL licenses state that all code files should mention the license, the conservative approach is to leave them in these reference files.

It's not important enough to be a blocking issue, so its best for me to remove my review.

yvaucher:
I merge this as if someone really want to put a copyright in such file he still can but having it in template module as recommendation make few sense. It just increase the number of edition an author of a module would have to do.

PR #271, opened 2017-04-11
improvement to pypi registration

sbidoul:
- sometimes the pypi registration process fails, so we retry twice before dying.
- cache registration status


dreispt:
@lmignon has a point. If not an actual parameter, at least it should be a CONSTANT.

sbidoul:
@dreispt @lmignon register retry count is now a constant

PR #272, opened 2017-05-12
[ADD] Demo record XML ID rule

obulkin:
* Add rule to OCA guidelines requiring all demo record XML IDs to end in `demo`
* Add rule to OCA guidelines requiring all XML IDs for data records to begin with the record's model name

The first portion was suggested [here](https://github.com/OCA/delivery-carrier/pull/122#discussion_r113841148) by @lasley and seems like a good move since it can be difficult to tell demo records apart from normal ones. The second part was then added based on [a discussion in this PR](https://github.com/OCA/maintainer-tools/pull/272#pullrequestreview-37844704).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -23,6 +23,7 @@ section.
       * [Format](#format)
       * [Records](#records)
       * [Naming xml_id](#naming-xml_id)
+        * [Data Records](#data-records)
         * [Security, View and Action](#security-view-and-action)
         * [Inherited XML](#inherited-xml)
         * [Demo Records](#demo-records)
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -389,6 +390,17 @@ high value in its `priority` (greater than 100 is recommended) to avoid the erro
...
 Also, we can hide an element from the view using `invisible="1"`.
 
+#### Demo Records
+
+Suffix all demo record XML IDs with `demo`. This allows them to be easily
+distinguished from regular records, which otherwise requires examining the
+source or reinstalling the module with demo data disabled.
+
+```xml
+<record id="not_a_real_user_demo" model="res.users">
```
lasley: You should still follow the standard naming schema though. 

```
model_name_something_demo
```

obulkin: Isn't that schema just for models with a `model` or `res_model` field, like views, actions, etc?
 It seems redundant in other cases.

lasley: It's only redundant in this context. It is not, however, when just seeing an external ID.

obulkin: IME, XML IDs generally show up in the context of record creation or assignment to a relational field, both of which make it fairly easy to tell what model is involved. 

That being said, I'm not opposed to this rule, but it's not part of the current guidelines as I understand them. The section on XML ID naming only has subsections on naming for specific models (menus, views, actions, groups, and rules) and on inheritance. Do you want to add this to the PR?

lasley: There are some situations in which you may only have an external ID - although I can't think of one off hand, it's definitely happened.

Hmmm you're right about there not being anything regarding data records in the guidelines though - both demo or non. Wouldn't hurt to expand the scope of this PR for all data records, instead of just Demo.

obulkin: Sure thing. I'll add that.

lasley:
@obulkin - please also update the template module to conform to these now accepted guidelines

obulkin:
@lasley - Sure thing. The template changes are here: https://github.com/OCA/maintainer-tools/pull/275

PR #275, opened 2017-05-23
[FIX] XML IDs in module template

obulkin:
* Update module template to reflect new XML ID rules
* Correct several XML IDs in module template not following older style rules

This PR is connected to the new rules introduced in https://github.com/OCA/maintainer-tools/pull/272

### DIFF
--- a/template/module/wizards/wizard_model_view.xml
+++ b/template/module/wizards/wizard_model_view.xml
@@ -5,7 +5,7 @@
 <openerp>
 <data>
 
-<record id="wizard_model_view" model="ir.ui.view">
+<record id="wizard_model_view_form" model="ir.ui.view">
     <field name="name">My view description</field>
     <field name="model">module.wizard_model</field>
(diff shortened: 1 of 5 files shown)


template/module/wizards/wizard_model_view.xml
```diff
@@ -5,7 +5,7 @@
 <openerp>
 <data>
 
-<record id="wizard_model_view" model="ir.ui.view">
+<record id="wizard_model_view_form" model="ir.ui.view">
```
pedrobaeza: This one should be also `model_name_view_form`. The model name determines if it's a transient model (wizard) or not.

lasley: @pedrobaeza - the model in this instance is `module.wizard_model`.

I think we should maybe revisit our wizard naming, although I dunno if it makes sense to do it here

pedrobaeza: OK then

PR #276, opened 2017-05-31
[IMP] new convention for website key

sbidoul:
I hereby propose a new convention for the website key in the manifest.

The current tradition is to put the author company website or the OCA website. Neither of these provide directly relevant information about the addon. 

I therefore propose to use the link to the OCA github repository which provides direct access to the addon README which in turn provides immediate access to the most relevant information, including detailed authors and contributors information.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -79,6 +79,10 @@ section.
   * Make sure it has the `license` and `images` keys.
   * Make sure the text `,Odoo Community Association (OCA)` is appended
     to the `author` text.
+  * The `website` key must be `https://github.com/OCA/<repo>/<addon>`,
+    so as to provide the most relevant link to discover more information about the addon.
+    That link shows the addon README which in turn shows proper credits (authors, contributors and theirs companies),
+    and links to the relevant information the OCA website.
 * Don't use your company logo or your corporate branding. Using the website, the author and the list of contributors is enough for people to know about your employer/company and contact you.
 
 ### Version numbers
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -84,7 +84,7 @@ section.
     That link shows the repository README. Alternatively `https://github.com/OCA/<repo>/tree/<branch>/<addon>`
     may be used, to provide a direct link to the addon README, which includes proper credits
     (authors, contributors and theirs companies), and links to the relevant information on the OCA website.
-* Don't use your company logo or your corporate branding. Using the website, the author and the list of contributors is enough for people to know about your employer/company and contact you.
+* Don't use your company logo or your corporate branding. Using the author and the list of contributors is enough for people to know about your employer/company and contact you.
 
 ### Version numbers
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -79,10 +79,11 @@ section.
   * Make sure it has the `license` and `images` keys.
   * Make sure the text `,Odoo Community Association (OCA)` is appended
     to the `author` text.
-  * The `website` key must be `https://github.com/OCA/<repo>/<addon>`,
+  * The `website` key must be `https://github.com/OCA/<repo>`,
     so as to provide the most relevant link to discover more information about the addon.
-    That link shows the addon README which in turn shows proper credits (authors, contributors and theirs companies),
-    and links to the relevant information the OCA website.
+    That link shows the repository README. Alternatively `https://github.com/OCA/<repo>/tree/<branch>/<addon>`
+    may be used, to provide a direct link to the addon README, which includes proper credits
+    (authors, contributors and theirs companies), and links to the relevant information on the OCA website.
 * Don't use your company logo or your corporate branding. Using the website, the author and the list of contributors is enough for people to know about your employer/company and contact you.
 
 ### Version numbers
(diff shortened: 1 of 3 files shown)


pedrobaeza:
Well, with this we are removing one of the few branding possibilities contributors have, and you know that some are reluctant because they can put logos and so on on other places. If we remove also this one, there's no way to go to any possible commercial page, and the only visible brand will be the name on the author list.

sbidoul:
@pedrobaeza I'd be happy to allow URLs in the authors/contributors section of the README. The website key is not very good for branding because it's only one URI, and so when they are multiple authors we fallback on odoo-community.org which is not informative either.

StefanRijnhart:
@sbidoul Agreed. This has always been my own practice when proposing new modules.

sbidoul:
@StefanRijnhart indeed, I did not double check my link.

Could we leave the freedom:
- `https://github.com/OCA/<repo>` for small, cohesive repos
- `https://github.com/OCA/<repo>/tree/<branch>/<addon>` for large multipurpose repos such as `server-tools`

sbidoul:
My objective is to have the link to the most relevant information, which can be either in the repository README for small repos with a well defined scope, or the addon README for addons living in a multipurpose repository.

StefanRijnhart:
Well, I think linking to the addon instead of the repo is actually less informative but it is a matter of taste but if you will allow linking to the repo itself as one of the options it is alright for me.

pedrobaeza:
Then let's also change the README template to include this branded links please

sbidoul:
@StefanRijnhart @pedrobaeza done

CONTRIBUTING.md
```diff
@@ -79,7 +79,12 @@ section.
   * Make sure it has the `license` and `images` keys.
   * Make sure the text `,Odoo Community Association (OCA)` is appended
     to the `author` text.
-* Don't use your company logo or your corporate branding. Using the website, the author and the list of contributors is enough for people to know about your employer/company and contact you.
+  * The `website` key must be `https://github.com/OCA/<repo>`,
```
dreispt: Should we prefer the more relevant `https://github.com/OCA/<repo>/tree/<branch>/` ?

template/module/README.rst
```diff
@@ -65,8 +65,8 @@ Images
 Contributors
 ------------
 
-* Firstname Lastname <email.address@example.org>
-* Second Person <second.person@example.org>
+* Firstname Lastname <email.address@example.org> (optional company website url)
```
elicoidal: A little strange to me to mix contributors (=usually individuals) with companies.
Could we think about separating or providing better visibility to the companies? Honestly funding and money comes most of the time from customers or integrators.

CONTRIBUTING.md
StefanRijnhart: Again, I myself prefer just the repo as it is the landing page of the whole project. So 'more relevant' is a matter of opinion likewise.

template/module/README.rst
sbidoul: We already have a [section for funders](https://github.com/OCA/maintainer-tools/blob/d33158faa1b762a7428a446a6f9711c419546547/template/module/README.rst#funders).

Often contributors are individuals paid by their company so putting an URL to the company web site next to contributor emails sounds reasonable.

CONTRIBUTING.md
sbidoul: Should I rework the wording to give the same weight to both options? Something like `The website key must be A or B, whichever is more informative.`

sbidoul:
Hi all, anymore comments on this proposal? Changes required? Alternatives?

template/module/README.rst
dreispt: I agree with @sbidouil - if I make a contribution an an employee, my employer should be shown alongside my name, since I'm acting on behalf of it.
But @elicoidal also has a point: the company funding the module development deserver some visibility. It would be nice for them to be allowed to display their logos in a section next (below?) to the contributor list.

pedrobaeza: I already added a section `Funders` for that (but for now without logo).

elicoidal:
I would be in favor of adding logo in funders.
Nevertheless currently rst format might result in not so nice output...

sbidoul:
If we allow logos, we may want to put size limits and best practice recommandations such as:
- maximum height 50px
- use a publicly accessible URL (not an image in the repo)
- provide an alt-text
- image format png or jpg

Here is the rst reference about images:
http://docutils.sourceforge.net/docs/ref/rst/directives.html#images

template/module/README.rst
dreispt: OK Pedro. So I guess this particular change is OK

elicoidal:
For logos, I would suggest something like this:
```
.. |logo1| image:: http://thomas-cokelaer.info/tutorials/sphinx/_images/stars.jpg
    :width: 50pt
    :height: 50pt

.. |logo2| image:: http://thomas-cokelaer.info/tutorials/sphinx/_images/stars.jpg
    :width: 50pt
    :height: 50pt


|logo1| |logo2|
```

moylop260:
@StefanRijnhart @pedrobaeza @elicoidal Do you have another comment?

pedrobaeza:
OK for me

PR #279, opened 2017-06-30
IMP better practice on library use

bealdav:
### Milestone (Odoo version)
- all

### Module(s)
- all, library use

### Fixes / new features
- better practice on library use

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -470,8 +470,9 @@ running tests and in such a case your module and their external dependencies are
 **You also need to test your code without the python lib nor the module installed** 
 to be sure other lines of your code don't break the installation.
 
+##### Recommandation
 A good practice is to only call library from a method or function. 
-Then code is only called when module is intalled.
+Then code is only called when module is installed.
 
 
 #### README
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -467,6 +467,13 @@ except (ImportError, IOError) as err:
 This rule doesn't apply to the test files since these files are loaded only when
 running tests and in such a case your module and their external dependencies are installed.
 
+**You also need to test your code without the python lib nor the module installed** 
+to be sure other lines of your code don't break the installation.
+
+A good practice is to only call library from a method or function. 
+Then code is only called when module is intalled.
+
+
 #### README
 If your module uses extra dependencies of python or binaries, please explain
 how to install them in the `README.rst` file in the section `Installation`.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -470,8 +470,8 @@ running tests and in such a case your module and their external dependencies are
 **You also need to test your code without the python lib nor the module installed** 
 to be sure other lines of your code don't break the installation.
 
-##### Recommandation
-A good practice is to only call library from a method or function. 
+##### Recommendation
+A good practice is to only call an external library from inside a method or function.
 Then code is only called when module is installed.
 
 


bealdav:
PSC can amend it. Thanks

CONTRIBUTING.md
```diff
@@ -467,6 +467,13 @@ except (ImportError, IOError) as err:
 This rule doesn't apply to the test files since these files are loaded only when
 running tests and in such a case your module and their external dependencies are installed.
 
+**You also need to test your code without the python lib nor the module installed** 
+to be sure other lines of your code don't break the installation.
+
+A good practice is to only call library from a method or function. 
+Then code is only called when module is intalled.
```
dreispt: installed

dreispt:
I wonder if this is practical.
Does the extra test effort compensate the possible benefits?
If so, a how to guideline would be welcome.

Also, the "you need" wording is not clear to me: is it intended to be a recommendation or mandatory/PR blocking?

CONTRIBUTING.md
```diff
@@ -467,6 +467,14 @@ except (ImportError, IOError) as err:
 This rule doesn't apply to the test files since these files are loaded only when
 running tests and in such a case your module and their external dependencies are installed.
 
+**You also need to test your code without the python lib nor the module installed** 
+to be sure other lines of your code don't break the installation.
+
+##### Recommandation
```
lasley: s/Recommandation/Recommendation

CONTRIBUTING.md
```diff
@@ -467,6 +467,14 @@ except (ImportError, IOError) as err:
 This rule doesn't apply to the test files since these files are loaded only when
 running tests and in such a case your module and their external dependencies are installed.
 
+**You also need to test your code without the python lib nor the module installed**
```
lasley: Like automated tests? How? Maybe this is something we can do with Pylint?

CONTRIBUTING.md
```diff
@@ -467,6 +467,14 @@ except (ImportError, IOError) as err:
 This rule doesn't apply to the test files since these files are loaded only when
 running tests and in such a case your module and their external dependencies are installed.
 
+**You also need to test your code without the python lib nor the module installed** 
+to be sure other lines of your code don't break the installation.
+
+##### Recommandation
+A good practice is to only call library from a method or function.
```
lasley: A good practice is to only call an external library from inside a method or function.

CONTRIBUTING.md
bealdav: For test: adding repo but without any installed any module but restart odoo

lasley: Hmmmm I think I might agree with @dreispt on the feasibility that this is actually going to be done - that's a decent amount of overhead to my less ops-oriented engineers.

Just so I'm making sure I understand the intent - it's to make sure that there are not any external library calls or whatever in the global namespace?

If this is the case, I feel like this is actually a case we can watch for in lint. @moylop260 do you have any thoughts on that?

moylop260: Currently we have a similar check from https://github.com/oca/pylint-odoo

 - W7935 | External dependency "%s" without ImportError. More info: https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#external-dependencies | missing-import-error
 - W7936 | Missing external dependency "%s" from manifest. More info: https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#external-dependencies | missing-manifest-dependency

If you use a `import` of a external library that odoo don't have defined in requirements pylint-odoo will emit a message to add a `try-except-ImportError` and a adds `external dependency` to manifest file

bealdav:
Here is https://github.com/OCA/server-tools/pull/883/files original PR which bring to this PR

moylop260:
That PR have [raven in the external_dependencies](https://github.com/versada/server-tools/blob/5098418edc9fbb1883244cea8c2086504777d2a3/sentry/__manifest__.py#L18)
if you try to install that module without the dependencies installed previously so you will have a raise error [Unable to install module "sentry" because an external dependency is not met: raven](https://github.com/odoo/odoo/blob/cbb52702dd07efe430eed8a76b3e2e39ce038b8a/odoo/addons/base/module/module.py#L312)

However, if you have that module installed before of the change to add `external_dependencies` and upgrade it (`git pull`) this error will be silent.

bealdav:
The problem was than the server raise even if module wasn't installed: this is the topic.

moylop260:
Or sorry, I understand your point now.
That module has [static folder](https://github.com/versada/server-tools/tree/5098418edc9fbb1883244cea8c2086504777d2a3/sentry/static) then odoo will compile it without install it and a variable defined from `try-variable-except` is used and you will raise a error.

Original pylint project has checks to other similar cases:
 - E0601 - used-before-assignment - Using variable %r before assignment
 - W0631 - undefined-loop-variable - Using possibly undefined loop variable %r

A good lint will be `Using possibly undefined try-except variable`

Could you create a RFC here: https://github.com/PyCQA/pylint/issues/new?

PR #280, opened 2017-07-02
[IMP] various improvements to pypi_upload_wheels

sbidoul:
- cope with the sunsetting of legacy pypi: https://mail.python.org/pipermail/distutils-sig/2017-June/030766.html
- `twine register` is not supported nor necessary with pypi.org
- use `dumbdbm` instead of `anydbm` for portability

@gurneyalex in addition to this I had to compile a python 2.7.13 on odoo-community.org, because wheezy's 2.7.3 is hopelessly obsolete wrt https support.


sbidoul:
This is working on the OCA server for several months now. I merge.

PR #290, opened 2017-09-11
[FIX] UICase Fixes

lasley:
* Add `post_install` and `at_install` keys
* Fix PEP

Not including the `post_install` and `at_install` keys can yield random errors, such as missing static QWeb templates.

Cc @hughesbm

### DIFF
--- a/template/module/tests/test_something.py
+++ b/template/module/tests/test_something.py
@@ -29,9 +29,17 @@ class SomethingCase(TransactionCase):
 
 
 class UICase(HttpCase):
+
+    post_install = True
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Shouldn't this be put as decorators?

lasley:
From what I can tell, it all works. [Both are used throughout Odoo code](https://github.com/odoo/odoo/search?utf8=%E2%9C%93&q=%22post_install%22&type=) - sometimes even both at the same time.

I chose this direction because it does not require an import.

PR #294, opened 2017-09-19
[IMP] contributing: restructurize guidelines

naglis:
### Milestone (Odoo version)
- None

### Module(s)
- None

### Fixes / new features
- Currently the section _External dependencies_ is under _XML files_ which feels like a non sequitur. I believe it makes more sense to put the section _External dependencies_ under _Modules_.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -19,6 +19,10 @@ section.
       * [Directories](#directories)
       * [File naming](#file-naming)
       * [Installation hooks](#installation-hooks)
+      * [External dependencies](#external-dependencies)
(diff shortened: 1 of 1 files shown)


pedrobaeza:
For reducing the diff, can we leave the section in the same place, but changing the TOC level?

naglis:
My first thought was to do that, but then it still felt a bit awkard when reading from top to bottom.

Is having the TOC not match the order of sections in the content OK? If so, I'll gladly update the PR.

pedrobaeza:
OK, I think it's OK to change it, but please take the occasion to change the section `__openerp__.py` to `Manifest (__manifest__/__openerp__.py)` to adapt it to v10.

naglis:
@lasley, sure, it is done.

lasley:
Thanks @naglis

PR #297, opened 2017-10-02
[ADD] sync for odoo 11

hbrunn:
this add sync from upstream for odoo 11. Only merge after the saas-18 branch is renamed.


pedrobaeza:
Labelling as WIP until then

hbrunn:
now with https://github.com/odoo/odoo/tree/11.0 existing, we can merge this (and shouldn't forget to update the script on the oca server that does the syncing)

sbidoul:
The cron on odoo-community.org is updated, let's see what happens tonight.

PR #300, opened 2017-10-03
migrate_branch_empty: New script for migrating empty branches

pedrobaeza:
It also includes an adaptation to version 11.0 for migrate_branch.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -30,6 +30,7 @@ console_scripts =
     oca-gen-addons-table = tools.gen_addons_table:main
     oca-gen-all-addons-tables = tools.gen_all_addons_tables:main
     oca-migrate-branch = tools.migrate_branch:main
+    oca-migrate-branch-empty = tools.migrate_branch_empty:main
     oca-bdist-all-wheels = tools.bdist_all_wheels:main
     oca-pypi-upload-wheels = tools.pypi_upload_wheels:cli
     oca-wh2simple = tools.wh2simple:main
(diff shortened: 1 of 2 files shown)


pedrobaeza:
This has been used on OCA 11.0 branch generation, so I think we can merge it.

StefanRijnhart:
Sorry, I missed this one. Thanks though ;-)

tools/migrate_branch.py
```diff
@@ -252,18 +252,37 @@ def _update_metafiles(self, repo, root_contents):
...
+                    ("\[//]: # \(addons\).*\[//]: # \(end addons\)",
+                     "[//]: # (addons)\n[//]: # (end addons)"),
+                ],
+            },
+            '.travis.yml': {
+                None: [
+                    (source_string, target_string),
+                    (source_string_dash, target_string_dash),
+                    (r"(?i)([^\n]+ODOO_REPO=['\"]ODOO[^\n]+)\n([^\n]+"
+                     r"ODOO_REPO=['\"]oca\/ocb[^\n]+)", r'\2\n\1'),
```
StefanRijnhart: BTW, this expression that we made, it may not give you the right results if there are more runs with INCLUDE/EXCLUDE right below each other. What was the reason for putting OCB first again?

pedrobaeza: It was for having runbot using OCB 11.0 branch for deployments instead of the missing odoo one. It doesn't make sense right now.

PR #302, opened 2017-10-07
gen_all_addons_tables 11.0

sbidoul:
As installed on the OCA server.

### DIFF
--- a/tools/gen_all_addons_tables.py
+++ b/tools/gen_all_addons_tables.py
@@ -14,7 +14,7 @@ import subprocess
 import sys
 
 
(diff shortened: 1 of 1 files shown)


pedrobaeza:
As 11.0 branches have been started empty, this script is not very needed right now.

sbidoul:
The table with addons versions and summaries is nice to have. Example https://github.com/OCA/queue

pedrobaeza:
Only mis_builder, connector and queue repositories have kept the installable = False method, but as this opt-in introducing in the README the section, OK for me.

Note that the rest of the repositories don't have right now that section, as I stripped it in the migration process.

PR #316, opened 2017-12-02
[ADD] gitignore Visual Studio .vs/

emagdalenaC2i:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,6 +48,9 @@ coverage.xml
 # Eclipse
 .settings
 
+# Visual Studio cache/options directory
+.vs/
+
 # Django stuff:
 *.log
 


emagdalenaC2i:
@zamberjo and @mreficent could you review this PR? It is more or less the same that the one merged a few days ago about Eclipse, but for Visual Studio

lasley:
@emagdalenaC2i Just confirming, this is meant as a repo level suggestion and not instead meant to be in the [MQT sample files](https://github.com/OCA/maintainer-quality-tools/blob/master/sample_files/.gitignore)?

emagdalenaC2i:
Well I think that the .gitignore file should be the same in all the OCA repos, and so... I think it should be later in the MQT sample files.

But for now, I just want to confirm that there is no problems with this PR. So I can create other PRs to change the .gitignore file in the repos I used to work (project and spain mainly)

PR #318, opened 2017-12-19
[ADD] gitignore: mac .DS_Store

mreficent:
I have seen some people using it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -51,6 +51,9 @@ coverage.xml
 # Visual Studio cache/options directory
 .vs/
 
+# OSX Files
+.DS_Store
+
 # Django stuff:
 *.log
 


mreficent:
@lasley, maybe. I think both files should have the same exclusions, or at least we should try that them have the maximum of exclusions in common possible.

lasley:
I agree @mreficent - consistency is key IMO. Can you please bring over the template here then?

mreficent:
@lasley comparing the two templates, what I see is that the **maintainer-tools** template (with this PR) includes all exclusions listed in the **maintainer-quality-tools** template. However, that template in **maintainer-quality-tools** is the incomplete one!

lasley:
Ah seems you're right @mreficent. I'll merge this one then.

Did you want to tackle updating the template in MQT for us as well?

PR #324, opened 2018-02-01
[ADD] a word about migrations

hbrunn:
### Milestone (Odoo version)
- all

### Module(s)
- all

### Fixes / new features
- conventions update

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -16,6 +16,7 @@ section.
   * [OCA Guidelines](#oca-guidelines)
     * [Modules](#modules)
       * [Version numbers](#version-numbers)
+      * [Migrations](#migrations)
       * [Directories](#directories)
       * [File naming](#file-naming)
       * [Installation hooks](#installation-hooks)
@@ -112,6 +113,9 @@ The `x.y.z` version numbers follow the semantics `breaking.feature.fix`:
 If applicable, breaking changes are expected to include instructions
 or scripts to perform migration on current installations.
 
+### Migrations
+
+When you introduce a breaking change, you *must* provide a migration script to make it possible to upgrade from lower versions. For a migration to another major version of Odoo, it's quite probable you'll need a migration script too. In such cases, migration scripts are highly appreciated, but a note in the README about relevant changes needing migration is sufficient too so that later contributors can add migration scripts without having to analyze all changes again.
 
 ### Directories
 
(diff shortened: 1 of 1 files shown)


sbidoul:
Hm. Stumbling on this one by chance. Merges go fast these days.

I agree with the part "when you introduce a breaking change".

For people migrating though, I'm afraid this introduces an additional hurdle to contribution and I think it's acceptable to decouple migration of the module from writing of the migration scripts.

hbrunn:
For migrations: Well, I see the point. But my thinking is that the person doing a migration had to go through all the changes anyways, so actually adding the migration script won't be so much extra work (provided the person in question is acquainted with migration scripts). And if the person is not acquainted with migration scripts, so much more reason to get acquainted, right?

I think for the ecosystem as a whole holes in the upgrade path are a quality problem.

pedrobaeza:
I totally agree with @hbrunn. And with this, people will be more sensible with some changes that are only aesthetic or not needed.

sbidoul:
I see your points. But imagine newcomers on v11 helping with migration. You are asking them to learn the data model of the previous version, install it, test the migration script... for something they have no direct use. Quite a deterrent, IMO.

simahawk:
I agree w/ @sbidoul . As this started from https://github.com/OCA/social/pull/225 I'd say that it would enough to provide proper info on what changed (see [my README](https://github.com/OCA/social/tree/11.0/mail_digest#migrating-to-v11)).
That will help everybody - even your future self - to add the mig step when really needed.

CONTRIBUTING.md
```diff
@@ -112,6 +112,9 @@ The `x.y.z` version numbers follow the semantics `breaking.feature.fix`:
 If applicable, breaking changes are expected to include instructions
 or scripts to perform migration on current installations.
 
+### Migrations
+
+When you you introduce a breaking change, and when you migrate a module to a new major version, you *must* provide a migration script to make it possible to upgrade from lower versions.
```
simahawk: that "must" sounds too much :)

hbrunn: well, for breaking changes within a version I'd insist on the script, so in RFC speech, that's a must I think

hbrunn:
can we agree on removing the major version in the above paragraph, but adding a sentence "For migrations to a new major version, migration scripts (if necessary) are highly appreciated, but a note about relevant changes needing migration is sufficient too" (I just don't want to lose the knowledge in the dev's head at this point in time, so what @simahawk did is fine with me too; I've merged it ;-)

CONTRIBUTING.md
hbrunn: https://www.ietf.org/rfc/rfc2119.txt

sbidoul:
@hbrunn I agree with your proposal.

hbrunn:
I made it a bit longer, that should make us all happy

PR #326, opened 2018-02-03
oca-copy-maintainers: manage PSC Representative team

gurneyalex:
At the request of the Board, I changed the script to handle a team for the PSC Representative of the projects. With this version, a team is created (if needed) for the PSC Representative with the proper access, and synchronised with the normal team, and the PSC Representative is added to the team. 

Result: 
image

### DIFF
--- a/tools/copy_maintainers.py
+++ b/tools/copy_maintainers.py
@@ -26,6 +26,7 @@ class FakeProject(object):
 
     @property
     def user_id(self):
+        return False
         return self._members[0] if self._members else False
 
     @property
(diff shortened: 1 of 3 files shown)


tools/copy_maintainers.py
```diff
@@ -26,6 +26,7 @@ def __init__(self, name, members):
 
     @property
     def user_id(self):
+        return False
```
pedrobaeza: Why this?

gurneyalex: user_id is the psc representative. We don't have one for the "virtual teams" such as OCA Member, for which this class is used. Could have hard coded Joël in there, but it did not seem useful

pedrobaeza: Thanks for the info. A comment here would help for future references.

PR #328, opened 2018-02-16
[RM] encoding=utf-8 is not needed anymore in Python 3

bodedra:
### Milestone (Odoo version)
- 11.0

### New features
- UTF-8 as the default source encoding for Python 3.x.

### DIFF
--- a/template/module/demo/assets.xml
+++ b/template/module/demo/assets.xml
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0"?>
 <!-- Copyright <YEAR(S)> <AUTHOR(S)>
      License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
 


pedrobaeza:
The problem is that this template is used for all versions...

bodedra:
May we create specific template for specific version ?

pedrobaeza:
I think we should ask for opinions on the contributors list

feketemihai:
@bodedra @pedrobaeza It's useful to have templates per version, but will be a lot of work on this

pedrobaeza:
Yeah, but right now, the template is for v9, not for v11. And if we change it, then other making modules for older versions will have the same problem.

elicoidal:
Actually maybe we should keep it as it is today: in the end having the coding line in P3 is not necessary but not harmful (AFAIK). In a next version, we could think of removing the line

dreispt:
I also think that it should not be removed, since it is still relevant for <=10.0.

mreficent:
What about changing this
```python
# -*- coding: utf-8 -*-
```
by this
```python
# -*- coding: utf-8 -*-  [remove this line in v11 or above]
```
?

pedrobaeza:
@bodedra can you do the latest suggestion?

bodedra:
@pedrobaeza Would you please guide me what I need to take care or what's your latest suggestion ?

leemannd:
@bodedra The goal is simply to do this =>  https://github.com/OCA/maintainer-tools/pull/328#issuecomment-384746221

mreficent:
@bodedra also please, delete last commit, because it's not good deleting the encoding in the xm files. @pedrobaeza  could you give a good explanation if you need :)

bodedra:
@leemannd that has been already taken care at initial commit.

Question: Do you want to revert back other changes and do only mentioned https://github.com/OCA/maintainer-tools/pull/328#issuecomment-384746221 ?

leemannd:
The last suggestion is to put:
`# -*- coding: utf-8 -*-  [remove this line in v11 or above]`
It is the only part that really matter right?

pedrobaeza:
@bodedra any update on this?

bodedra:
@pedrobaeza I have tried to put it back. Somehow travis failed. Would you please help me out here to understand what cause ?

@leemannd @mreficent

pedrobaeza:
Not sure... @moylop260 do you have a clue?

moylop260:
127 commits are fine?
Should you rebase?

template/module/demo/assets.xml
```diff
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0"?>
```
mreficent: Wrong, just modify `.py` files

bodedra:
Hello Team,

I have created a new PR https://github.com/OCA/maintainer-tools/pull/437 and make it clean.

PR #333, opened 2018-04-10
Remove license from __init__.py

elicoidal:
Usually license and authorship are not to be added in __init__.py as they are not supposed to contain code.


elicoidal:
@pedrobaeza @dreispt @moylop260

pedrobaeza:
It's not the same the copyright information than the license one IMO.

dreispt:
AFAIK the common agreement is that it is optional, but some people take the license text to the letter, and it says the copyright notice should be in all code files.
IMO we shouldn't remove it from templates.

PR #336, opened 2018-04-17
Easier mechanism for scripts to work on clones of OCA repos and improvements to packaging

sbidoul:
With application to the addons table generator.

Will soon be used on the new wheel/sdist builder.

This is the version that runs now on the OCA server.

This mechanism is simpler to use and more robust that doing oca-clone-everything before processing. In particular this reduce the chance of push conflict because the time between clone and push is reduced.


sbidoul:
This is running fine on the OCA server. No objection to merging?

PR #337, opened 2018-04-18
Add ES logs instructions.

Yajo:
@Tecnativa

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -317,7 +317,7 @@ how to install them in the `README.rst` file in the section `Installation`.
 #### requirements.txt
 
 As specified in [Repositories](#repositories), you should also define
-the python packages to install in a file `requirements.txt` in the 
+the python packages to install in a file `requirements.txt` in the
 root folder of the repository. This will be used for travis.
 
 ## XML files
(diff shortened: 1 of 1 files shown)


pedrobaeza:
@mart-e what do you think about this?

mart-e:
You mean to integrate it in the documentation upstream?
Sometimes, I am a bit pessimistic about this. Usually, the people who make the mistakes are the one who do _not_ read the documentation and help.
Well, at least giving a section we can link to could be useful for maintainers.
A section in https://github.com/odoo/odoo/wiki/Contributing ?

pedrobaeza:
Yes, having the link you don't need to copy/paste each time the same text, and you can shut certain mouths saying "This is not explained anywhere"...

Yajo:
I'm personally the PSC of web and website repos in OCA; and it's way very common to have issues with this problem.

Even python programmers which don't understand JS make this failure, not only users. I think it would be very helpful.

Thanks!

mart-e:
Something like this? :wink: 
https://github.com/odoo/odoo/wiki/Contributing#tracebacks
(don't hesitate to correct if needed, I believe you have the access to do so Pedro)

Yajo:
I think this part shouldn't be a code block:

> In Odoo >= 10.0: Go to `Settings > Activate the developer mode (with assets)`.
In Odoo < 10.0: Click on your user menu (top right corner) `> About > Activate the developer mode`.

Other than that, thank you very much!! closing

pedrobaeza:
Great, thanks!

PR #339, opened 2018-05-05
README.rst generator for addons

### LINKED ISSUE #335: [RFC] README.rst generator
This follows #295 and a discussion with @elicoidal while working on the mis builder documentation.

Currently the README.rst is created manually by module authors. It contains 
- a mix of hand-written information (description, usage, credits, etc), 
- boiler plate OCA text (eg reference to OCA maintainers, link to runbot, link to issue tracker),
- some information that is redundant with the manifest (eg license badge)

The current situation makes it difficult to maintain consistency across OCA addons in the boilerplate text. It also makes it difficult to ensure consistency between the manifest and the README (eg for the license badge and the upcoming development status).

This is a proposal to let authors write the relevant text in small files and have the README.rst be generated nightly.

# Proposal

Let authors create reStructuredText fragments in the addon root directory,:
- introduction.rst
- installation.rst (optional)
- configuration.rst (optional)
- usage.rst
- developer.rst (optional, developer instructions)
- roadmap.rst (optional, known issues, future ideas)
- contributors.rst (optional)
- credits.rst (optional, additional credits other than authors and contributors)
- history.rst (optional, the changelog)

Each rst fragment should be composed of body text (paragraph, lists, tables, images) and *must not* contain any section title.

Create a new script `oca-gen-addon-readme` that generates README.rst for each addon in a repo where introduction.rst and usage.rst are present. The sections content is drawn from the rst files. Section titles are not generated if the corresponding rst file is absent. Title, license badge and development status badge are obtained from the addon manifest. If introduction.rst or usage.rst is missing, README.rst is left unchanged.

Update the nightly job that updates repos README.md to invoke `oca-gen-addon-readme` too.

sbidoul:
This implements and closes #335 

An example usage and resulting README.rst can be found in https://github.com/OCA/mis-builder/pull/90

TODO

- [x] support more licenses
- [x] support `maintainers` manifest key
- [x] `--addons-dir` (for multiple addons at once)
- [x] `--commit` option
- [ ] nightly run on all repos
- [x] update module template
- [x] link to a page defining the maintainer role ([WIP](https://drive.google.com/open?id=1ewkGkHzaKgtvnZZhp0ugSHiKZhlo3lechmGwsJ4OdNc))
- [x] maturity level badges must link to a page defining each level ([WIP](https://drive.google.com/open?id=1MmGxv0yTw0pvB-PYAcUhbCcku9ZG-Y2DSGXwZDNYyIw))
- [x] implement default development_status (Beta, in current state of discussions)


tools/gen_addon_readme.template
```diff
@@ -0,0 +1,86 @@
...
+{{- fragment('DEVELOP', 'Development') }}
+{{- fragment('ROADMAP', 'Known issues / Roadmap') -}}
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues <https://github.com/OCA/{{ repo_name }}/issues>`_.
+In case of trouble, please check there if your issue has already been reported.
+If you spotted it first, help us smashing it by providing a detailed and welcomed feedback.
+
+Do not contact contributors directly about support or help with technical issues.
```
dreispt: Probably out of scope for this PR, but it would be nice to add some actionable advice here. How about: "Either open an Issue on the Github project, or post your query to the OCA Contributors mailing list.". Not sure if we have a link to the Github project anywhere in the readme; we surley should.

tools/gen_addon_readme.template
```diff
@@ -0,0 +1,86 @@
...
+Current maintainer{% if manifest.maintainers|length > 1 %}s{% endif %}: 
+
+{% for maintainer in manifest.maintainers %}|maintainer-{{ maintainer }}| {% endfor -%}
+{% endif %}
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose
+mission is to support the collaborative development of Odoo features and
+promote its widespread use.
+
+To contribute to this module, please visit https://odoo-community.org.
```
dreispt: IMO this link is quite vague for the goal proposed.
Perhaps say "You are welcome to contribute. To learn how please visit https://odoo-community.org/page/Contribute".

template/module/INSTALL.rst
```diff
@@ -0,0 +1,7 @@
+[ this file must only be present if there are very specific
+  installation instructions, such as installing non-python
+  dependencies; the audience is systems administrators  ]
+
+To install this module, you need to:
+
+#. Do this ...
```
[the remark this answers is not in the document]
sbidoul: Well I'm not sure yet. Since INSTALL and CONFIGURE are for different audiences (system administrators and advanced users respectively), they can end-up in completely different place in a more complete documentation, so we should be able to include them separately in that larger book. Same reason why I kept ROADMAP and DEVELOP separate for now.

tools/gen_addon_readme.template
sbidoul: Good idea. I was also thinking to add a link to the repo here.

tools/gen_addon_readme.template
sbidoul: The repo link is 4 lines above.

tools/manifest.py
```diff
@@ -0,0 +1,43 @@
...
+
+def get_manifest_path(addon_dir):
+    for manifest_name in MANIFEST_NAMES:
+        manifest_path = os.path.join(addon_dir, manifest_name)
+        if os.path.isfile(manifest_path):
+            return manifest_path
+
+
+def parse_manifest(s):
+    return ast.literal_eval(s)
```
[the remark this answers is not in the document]
sbidoul: Why mature? I was proposing alpha as default, so maintainers can propose the level they consider appropriate with PR's and let the PSC approve that on a case by case basis. In any case we need to lay down the criteria before pushing this.

sbidoul:
@dreispt thanks for reviewing.

IMO it's important to have a clear and published definition for the maintainer role and the maturity levels before deploying this. At least basic criteria, otherwise confusion will ensue.

Also, if there are floating ideas to improve the README structure and/or boilerplate, let's implement that in this PR. Because each modification to a README will create a commit which in turns triggers a rebuild and publishing of new version to pypi of all addons. It's not dramatic, but we should strive for some stability.

template/module/INSTALL.rst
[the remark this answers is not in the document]
elicoidal: In general better to keep the options open for granularity.
We are talking about 10 files in total: it looks reasonable

tools/manifest.py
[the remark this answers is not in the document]
elicoidal: Agree: default should be the lowest: the contributor has to learn the way up

pedrobaeza:
The problem with the maturity is that the developer shouldn't be the one that assigns this level: this is for what we are creating the release cycle, and should be computed somehow.

sbidoul:
I think the maturity must be set by the developer and then CI checks that the corresponding criteria are satisfied. That is for criteria that can be checked automatically, because we may have some criteria that are not checkable by a tools, and can only be verified with human review.

tools/manifest.py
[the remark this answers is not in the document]
dreispt: I'm sorry, the default should rather be "Production/Stable",
The existing policy is that merged modules are Stable.
So, all current modules should be displayed as Stable, and we don't need to do a mass change to all modules.
New modules to be merged as Alpha or Beta should carry that in their manifest. This could be CI checked.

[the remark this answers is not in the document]
sbidoul: > The existing policy is that merged modules are Stable.

Let's finalize the criteria and then see if that holds true for all existing modules. In which case we can make a script to add it to existing modules. But it must not be the default, as the default would apply to all new modules.

dreispt:
I created the page explaining the Maturity levels: https://odoo-community.org/page/module-maturity-levels

elicoidal:
IMO: maturity should be maintained in separate environment (eg: ERP) where an human ("auditor") should be allowed to change/validate from one level to another: it has to be a manual check at some point (whether the developer sets it up or not).
 
Side note: I do not think all addons currently in OCA repos are "Stable" grade and we should probably let the PSC maintainer propose list of modules from the repo they are responsible of and the level accordingly.

sbidoul:
@dreispt thanks a lot! Would you put it first in a google doc to facilitate final review?

@elicoidal I agree with the manual check, but it can be done in github PR. We had that discussion before and we concluded we could start without storing the maturity level outside of git.

I also agree with a human intervention for setting the initial maturity levels.

dreispt:
Do you have any examples of OCA modules that don't follow Stable requirements?

dreispt:
The Maturity description is basically what was already [discussed](https://docs.google.com/document/d/1wwnu7oe5cDTyjZH3hdKkGLnF6fH4jPP3RLMPAhm3gjs/edit), but it could benefit from further fine tuning.
I've transcribed it to a [Google Doc](https://docs.google.com/document/d/1MmGxv0yTw0pvB-PYAcUhbCcku9ZG-Y2DSGXwZDNYyIw/edit?usp=sharing)
I think we should open a thread on the ML to gather more feedback on it.

sbidoul:
As an experiment I added color cues on the maturity level badges. As well as a github badge for quick access to the repo. Examples: [mis_builder](https://github.com/acsone/mis-builder/blob/fef84c70f170f42788af8d027768e8b621799ec2/mis_builder/README.rst) (stable), [mis_builder_demo](https://github.com/acsone/mis-builder/blob/fef84c70f170f42788af8d027768e8b621799ec2/mis_builder_demo/README.rst) (beta).

The more I think of it the more I think this shields.io badge mechanism is the most flexible while providing a decent enough look and feel. cc/ @max3903

elicoidal:
@sbidoul I like it but the maturity badge which is important is not highlighted in this situation while it is a very important one. That is my only concern as the look and feel is indeed good

sbidoul:
@elicoidal shields.io has an option  to make a badge bigger. OTOH with the color cue (ie Beta = orange) the important part of the maturity information is conveyed efficiently IMO.

sbidoul:
We could also display the maturity badge leftmost, before the license.

pedrobaeza:
Maybe in a different line even with a title?

sbidoul:
Example with [beta maturity badge on the left.](https://github.com/acsone/mis-builder/blob/c56ea7e05acfda20fd807580617bbd4956bb7584/mis_builder_demo/README.rst)

pedrobaeza:
What I'm missing here is the website URL of the authors, but I don't see easy way to add it with current boiler plate.

bealdav:
Just an opinion, but don't you think, all the meta data files with capitalized names such CONTRIBUTORS.rst could be to better place in a subfolder ? 
I don't know which name we should use ? metadata ?

sbidoul:
@bealdav yes I also considered that. Lacking a better idea I chose to keep them in the root folder. Which is not that exotic, we find such files in many projects. We have just a few more here.

Note it must not be `doc` nor `docs` which are traditionally reserved for Sphinx. And Sphinx wants all rst files to be in the toctree so that would conflict.

pedrobaeza:
What do we do with the authors websites?

sbidoul:
For the author web sites, you have the possibility to add urls in [CONTRIBUTORS.rst](https://github.com/acsone/maintainer-tools/blob/gen-addon-readme/template/module/CONTRIBUTORS.rst).

Alternatively we can chose to remove the Author section from the README (it currently comes from the manifest author key), and let people write a bit more in CONTRIBUTORS.rst.

template/module/README.rst
```diff
@@ -1,94 +1,20 @@
...
-Installation
-============
+* **DESCRIPTION.rst** (required)
+* INSTALL.rst (optional)
+* CONFIGURE.rst (optional)
+* **USAGE.rst** (optional, highly recommended)
+* DEVELOP.rst (optional)
+* ROADMAP.rst (optional)
+* **CONTRIBUTORS.rst** (optional, highly recommended)
+* CREDITS.rst (optional)
```
StefanRijnhart: HISTORY.rst (optional)?

sbidoul: So far I did not include the changelog in the readme because I feared it would be too long. But we could. What do you prefer?

sbidoul: The audience for a change log is users (bug fixes, improvements). A git commit history is too noisy for users (even developers using a library). There is always a lot of commits that do not belong to a changelog (eg refactorings), but are interesting for developers nevertheless. Moreover having several commits for a single features is often a desirable thing that also creates too much "noise" for a changelog.

So HISTORY.rst and the commit history are definitely two different things.

I would not add the link to the commit history in the README (the link to the repo is already a huge improvement and devs can easily find commits themselves).

I would love if we can gently push maintainers to start creating user oriented changelogs. That's why I also suggested to make it mandatory for mature addons.

bealdav:
@sbidoul About folder for FILE.rst, I suggest `readme` as folder name. It becomes explicit, I think

template/module/README.rst
lmignon: What about using github issues and milestones to manage the history? This would allow us to reference in the commit messages the context in which these changes occurs. From an issue we would also have the list of changes made to solve the problem. Thinking about it there are cases where issue and PR could be redundant and this would force us to create issues for different PR. This may not be the best idea, especially since this system will not work with our modules where we have 1 master branch per Odoo version. 
So, IMO, the best way to properly and efficiently manage the history is to have a HISTORY.rst file.

StefanRijnhart: OK, but if they are user oriented then you could include them in the README.rst, otherwise nobody will look at them.

sbidoul: @StefanRijnhart @lmignon thanks for your input. I updated the generator to include HISTORY.rst. See the updated [example](https://github.com/acsone/mis-builder/blob/10.0-generated-readme/mis_builder/README.rst).

sbidoul:
About a `readme` folder vs putting the fragments in the root folder, I'm neutral on this. What do people think?

lmignon:
@sbidoul About 'readme' It seems to me that having all these files at root is a common practice in a lot of python project.

template/module/README.rst
lmignon: @sbidoul Wouldn't it be more relevant to have the history at the end of the readme. It can become very long and reject the last two sections outside the reader's view.

sbidoul: @lmignon could be. My reasoning was that HISTORY (past) and ROADME (future) fit well next to each other. Also having the even more important maintainers section at then makes it easily accessible by scrolling to the end.

lmignon: @sbidoul :+1: that makes sense...

tools/manifest.py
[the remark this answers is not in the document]
yvaucher: +1 @sbidoul 
I'm also in favor of a validation from the developer that his module his mature instead of the way around.

Thus keeping Alpha as default.

elicoidal:
I like fragments in `readme` directory and README.rst in the root

tools/manifest.py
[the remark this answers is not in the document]
sbidoul: @yvaucher thanks for your input. In case you have not seen it the criteria for the levels are being discussed in a [google doc](https://docs.google.com/document/d/1MmGxv0yTw0pvB-PYAcUhbCcku9ZG-Y2DSGXwZDNYyIw/edit#). In that document the current proposed default is Beta.

mourad-ehm:
I agree with @elicoidal  because ReadMe files is not determinant files in a module (Python project). Put all fragments in readme directory will increase readability of code.

sebastienbeau:
I really prefer to have one directory with the all fragment, and then the full-readme at the racine.
+ 1000 for this magic tools !

sbidoul:
Ok, the vote is clear, thanks to all who participated. 

So latest changes:
- `readme` directory for the fragments
- validation of the generated readme
- default `development_status` is `Beta`

template/module/__openerp__.py
```diff
@@ -5,9 +5,11 @@
     "name": "Module name",
     "summary": "Module summary",
     "version": "8.0.1.0.0",
```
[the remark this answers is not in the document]
dreispt: IMO that is out of scope for this PR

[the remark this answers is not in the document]
sbidoul: @mourad-ehm out of scope for this PR indeed, but trival so I did it.

sbidoul:
There is one last open topic to resolve, then we can move forward.

Should HISTORY.rst (a user oriented changelog) be mandatory for Mature modules, or merely recommended. I'm personally in favor, since it is a benefits for users and integrators in need to decide to upgrade a module or not.

pedrobaeza:
Well, it's not blocking for this PR, but you know that my vote is no.

StefanRijnhart:
Prefer not mandatory

elicoidal:
prefer not mandatory

bealdav:
as said by @pedrobaeza we can merge as it and debates on history later

Thanks a lot for this really expected feature

sbidoul:
Ok, I added the links to the development status description page, as well as the maintainer role description (currently with a redirect to the google docs). I'll setup the nightly run soon. 

This PR is now complete. Thanks to all who commented.

pedrobaeza:
Should I squash commit history on merge?

sbidoul:
@pedrobaeza I prefer not squashing. The history is relatively clean and there are many commit that have a lot sense remaining standalone. I can try to squash some later during the week if you really want to.

pedrobaeza:
Yes, please, 48 commits is a lot for a script and you have 1 line = 1 commit for most of them.

sbidoul:
squished and squashed

sbidoul:
and deployed on OCA server

dreispt:
🎉

sbidoul:
Good. @dreispt  Now we need to finish and publish the maintainer roles and development status documents. Then communicate.

PR #342, opened 2018-05-24
Some housekeeping in OCA/maintainer-tools

sbidoul:
The list of commits should explain what I'm doing here


setup.py
```diff
@@ -1,10 +1,76 @@
+import os
```
elicoidal: I would suggest to take the opportunity to add headers for copyright and license to all files.

sbidoul:
License headers added.

sbidoul:
No objection if I merge this?

pedrobaeza:
OK, go ahead

PR #345, opened 2018-06-06
[ADD] Field Service

max3903:


max3903:
@pedrobaeza Field service is not a vertical. It's a functional area that can be found in many different industries...

pedrobaeza:
I don't agree. You can have modules that complement sales, purchase, rma, etc specially for fsm.

max3903:
I agree, but the field service team/department is a separate team than sales, purchase or rma. Field service activities have an impact on other functional areas as their processes are integrated with the others.

You can have a field service team/dept in an IT consulting company, in a construction company, in an ISP, in a hospital, in a manufacturing company, in any organization that sends someone on site/to an external facility to do something.

pedrobaeza:
Then you are giving me the reason for calling this repository with `vertical-`. Maybe your modules are for only the field service team, but there can be a lot more than are integrations with the rest of the system.

max3903:
@pedrobaeza I think you misunderstood me. The field service repo will not host sales, purchase or rma modules... only glue modules to make the bridge with those functional areas. 

You would not rename `purchase` to `vertical-purchase`, right? Same with field-service.

pedrobaeza:
No, you are the one that aren't understanding me, but this is the chicken-egg debate. We should vote about this @OCA/board what do you think?

dreispt:
I don't see why this should be a board restricted discussion.
It should be discussed in the Project & Service ML, or in the Contributors ML.

Issue #203, opened 2016-07-12
OCA Github Bot


jgrandguillaume:
[![Bountysource](https://api.bountysource.com/badge/issue?issue_id=35954912)](https://www.bountysource.com/issues/35954912-oca-github-bot?utm_source=35954912&utm_medium=shield&utm_campaign=ISSUE_BADGE)

# Goal
- Develop a bot that will help the review process for the maintainer, help newcomers, close too old PRs, etc..

# Current troubles
- Time to merge, # of PR increases 
  - → Generates frustration among new contributors
  - → Discouraging contributors
- Travis is not always green
  - Lots of time taken to make Travis green
  - Not easy for newcomers to make it green
- Merging rules might be too strict ?
- PSC might have the right to merge faster? Yes, PSC can decide if merge something although some guidelines are not met.
- Optimistic Merging (OM) instead of Pessimistic Merging (PM)?
- Difference between new modules and existing modules ?
- Different maturity levels in the same projects/repo in term of quality, reliability, usage
- Different level of maintenance among them (some are very active, others abandoned)
- Different levels of responsibility from the PSC

# Wanted Features

## Version 1
- Tag pull requests with "New contributor" for people with less than 3 PR within the OCA repos
- Close and comment ("Please re-open if necessary") every PR older than 6 months without comments
- Tag and comment if the pull request is ready to be merged (2 approvals, 5 days, green CI) 

## Later
- Tag “new module proposal” or “improvements on existing”
- Post message in case of:
  - Travis red -> how to solve
  - Newcomers -> Say hello, thanks you, where to look for help + pointer to doc or relevant link

## Resources
Use : https://github.com/facebook/mention-bot
Note this module already interact with github and odoo https://github.com/legalsylvain/oca-custom

pedrobaeza:
Add also this one:

* Announce that the PR is ready to be merged if there are 2 approvals, not current comments and there are more than 5 days from the initial request.

hbrunn:
yes, @pedrobaeza's is a very good one. Even better would be if the bot set a label for that, then we can filter for that and merge with just a glance on the comments if there's anything blocking

pedrobaeza:
Yeah, tagging + mentioning should be the perfect combination!

lasley:
> Close and comment ("Please re-open if necessary") every PR older than 6 months without comments

[...]

PR #346, opened 2018-06-07
[ADD] Pull request review script

madman-bob:
Add a script to automatically review pull requests in the manner described in #203.

### DIFF
--- a/tools/review_pull_request.py
+++ b/tools/review_pull_request.py
@@ -1,0 +1,167 @@
+#!/usr/bin/env python
+# License AGPLv3 (http://www.gnu.org/licenses/agpl-3.0-standalone.html)
+
+import argparse
+from collections import Counter
+from datetime import datetime, timedelta
+
+from github_login import login
+
+from oca_projects import OCA_USERNAME, OCA_REPOSITORY_NAMES
+
+contributions_cache = None
+
+
+def get_contribution_counts():
+    global contributions_cache
+
+    if contributions_cache is not None:
+        return contributions_cache
+
+    github = login()
+    contributions_cache = Counter()
+
+    for repository_name in OCA_REPOSITORY_NAMES:
+        repository = github.repository(OCA_USERNAME, repository_name)
+        for user in repository.contributors():
+            contributions_cache[user.id] += user.contributions_count
+
+    return contributions_cache
+
+
+def tag_new_pull_requests(repository_name):
+    github = login()
+    repository = github.repository(OCA_USERNAME, repository_name)
+    contributions = get_contribution_counts()
+
+    for pull_request in repository.pull_requests():
+        if contributions[pull_request.user.id] < 3:
+            pull_request.issue().add_labels('new contributor')
+
+
+def tag_all_new_pull_requests():
+    for repository_name in OCA_REPOSITORY_NAMES:
+        tag_new_pull_requests(repository_name)
+
+
+def close_abandoned_pull_requests(repository_name):
+    github = login()
+    repository = github.repository(OCA_USERNAME, repository_name)
+
+    for pull_request in repository.pull_requests():
+        full_pull_request = repository.pull_request(pull_request.number)
+
+        now = datetime.now(pull_request.created_at.tzinfo)
+        pr_is_old = now - pull_request.created_at > timedelta(days=31 * 6)
+
+        if pr_is_old and full_pull_request.comments_count == 0:
+            pull_request.create_comment("Please re-open if necessary")
+            pull_request.close()
+
+
+def close_all_abandoned_pull_requests():
+    for repository_name in OCA_REPOSITORY_NAMES:
+        close_abandoned_pull_requests(repository_name)
+
+
+def tag_ready_pull_requests(repository_name):
+    github = login()
+    repository = github.repository(OCA_USERNAME, repository_name)
+
+    for pull_request in repository.pull_requests():
+        full_pull_request = repository.pull_request(pull_request.number)
+
+        has_enough_reviews = sum(
+            1
+            for review in pull_request.reviews()
+            if review.state == 'APPROVED'
+        ) >= 2
+
+        now = datetime.now(pull_request.created_at.tzinfo)
+        pr_is_old_enough = now - pull_request.created_at > timedelta(days=5)
+
+        if (
+            has_enough_reviews and
+            pr_is_old_enough and
+            full_pull_request.mergeable
+        ):
+            pull_request.issue().add_labels('merge ready')
+            pull_request.create_comment("Ready to merge")
+
+
+def tag_all_ready_pull_requests():
+    for repository_name in OCA_REPOSITORY_NAMES:
+        tag_ready_pull_requests(repository_name)
+
+
+def main():
+    parser = argparse.ArgumentParser()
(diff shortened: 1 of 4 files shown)


jgrandguillaume:
Hi @madman-bob 

Thanks for your proposal ! On my side it looks good. I have one request for testing/fixing purpose:

Can you add a --dry-run option that will print the results instead of making them ?

Other than that, I've asked my colleague to review it as well.

Regards,

Joël

tools/review_pull_request.py
```diff
@@ -0,0 +1,167 @@
...
+    repository = github.repository(OCA_USERNAME, repository_name)
+
+    for pull_request in repository.pull_requests():
+        full_pull_request = repository.pull_request(pull_request.number)
+
+        now = datetime.now(pull_request.created_at.tzinfo)
+        pr_is_old = now - pull_request.created_at > timedelta(days=31 * 6)
+
+        if pr_is_old and full_pull_request.comments_count == 0:
+            pull_request.create_comment("Please re-open if necessary")
```
pedrobaeza: We can put here a longer text for not being rude, like:
```
This is an automatic message for telling you that this PR is going to be closed due to a long inactivity period.

Please ask for reopening if needed.
"""

File: tools/review_pull_request.py [hunk 2]
```diff
@@ -0,0 +1,167 @@
...
+    github = login()
+    repository = github.repository(OCA_USERNAME, repository_name)
+
+    for pull_request in repository.pull_requests():
+        full_pull_request = repository.pull_request(pull_request.number)
+
+        now = datetime.now(pull_request.created_at.tzinfo)
+        pr_is_old = now - pull_request.created_at > timedelta(days=31 * 6)
+
+        if pr_is_old and full_pull_request.comments_count == 0:
```
pedrobaeza: We can have PRs with comments, but also abandoned. I think a better metric that the number of comments is the date of the PR < 6 months if no comment, but the date of the last comment < 6 months in case that you have comments.

tools/review_pull_request.py
```diff
@@ -0,0 +1,167 @@
...
+
+def close_abandoned_pull_requests(repository_name):
+    github = login()
+    repository = github.repository(OCA_USERNAME, repository_name)
+
+    for pull_request in repository.pull_requests():
+        full_pull_request = repository.pull_request(pull_request.number)
+
+        now = datetime.now(pull_request.created_at.tzinfo)
+        pr_is_old = now - pull_request.created_at > timedelta(days=31 * 6)
```
pedrobaeza: Put directly 365 / 2 ~= 183

pedrobaeza:
I'm thinking right now about the GitHub rate limits, that can be a problem making this process. Look how I handled these limits when pulling Transifex translations. I wrap any GitHub call this way:

https://github.com/OCA/maintainer-tools/blob/20557facbd0c1eb40c843085b78c9f80d6ed2bd2/tools/tx_pull.py#L137-L166

and call it this way:

https://github.com/OCA/maintainer-tools/blob/20557facbd0c1eb40c843085b78c9f80d6ed2bd2/tools/tx_pull.py#L277

tools/review_pull_request.py
```diff
@@ -0,0 +1,167 @@
+#!/usr/bin/env python
```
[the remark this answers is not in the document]
madman-bob: Tested in 2.7. Works in 3.6 with a minor change to config.py. Not using any particularly exciting language features, so should work in all 3.x.

tools/review_pull_request.py
madman-bob: I was just following the requirements as lain out in the original issue. You may well be right, but I don't really want to get into discussing what the best metrics are. It should be easy enough to change, and feel free to do so, if you want to.

pedrobaeza: Yeah, thanks, but as you know the API, please do it you. I give you my approval on behalf the @OCA/board

tools/review_pull_request.py
lreficent: Just a nitpicking comment here: isn't better to use the last update date? If some big PR is having a long discussion it will be auto-closed after 6 months even if there is recent conversations/reviews.

lreficent: Then the comments_count could be removed

tools/review_pull_request.py
```diff
@@ -0,0 +1,221 @@
...
+def tag_new_pull_requests(repository_name, dry_run=False):
+    github = login()
+    repository = wrap_github_call(
+        github.repository,
+        [OCA_USERNAME, repository_name]
+    )
+    contributions = get_contribution_counts()
+
+    for pull_request in wrap_github_call(repository.pull_requests):
+        if contributions[pull_request.user.id] < 3:
```
hbrunn: I think this (and all other hardcoded values further down) should be arguments with the curernt value as default value

tools/review_pull_request.py
```diff
@@ -0,0 +1,221 @@
...
+    )
+
+    for pull_request in wrap_github_call(repository.pull_requests):
+        full_pull_request = wrap_github_call(
+            repository.pull_request,
+            [pull_request.number]
+        )
+
+        now = datetime.now(pull_request.created_at.tzinfo)
+        pr_is_old = now - pull_request.created_at > timedelta(days=31 * 6)
```
hbrunn: same thing as above

tools/review_pull_request.py
```diff
@@ -0,0 +1,221 @@
...
+             "instead of applying them.",
+        action="store_true"
+    )
+    args = parser.parse_args()
+
+    for repository_name in args.tag_new_pull_requests:
+        tag_new_pull_requests(repository_name, dry_run=args.dry_run)
+
+    if args.tag_all_new_pull_requests:
+        tag_all_new_pull_requests(dry_run=args.dry_run)
```
hbrunn: if you pass `args` to the functions, it's much simpler to implement my proposals above

tools/review_pull_request.py
```diff
@@ -0,0 +1,221 @@
...
+    return contributions_cache
+
+
+def tag_new_pull_requests(repository_name, dry_run=False):
+    github = login()
+    repository = wrap_github_call(
+        github.repository,
+        [OCA_USERNAME, repository_name]
+    )
+    contributions = get_contribution_counts()
```
sbidoul: I have concerns about the performance of this.

madman-bob: In what way?

madman-bob:
Have rebased onto `master`, to resolve merge conflicts.

max3903:
@madman-bob Thank you for your contribution. The @OCA/board will be testing your script in the next few months so the merge may only happen at the OCA code sprint in Belgium in October. If you want to claim the bounty earlier, please start the process (I think you have to initiate the process).

madman-bob:
@max3903 Thanks, but I'm only allowed to start the bounty claim once the original issue has been closed.

max3903:
@madman-bob Done

madman-bob:
@max3903 Thank you very much. Claim submitted.

pedrobaeza:
@sbidoul can this be closed as it has been moved to specialized repo?

sbidoul:
Yes!

PR #349, opened 2018-06-14
add: tool to configure travis for an OCA repo

sbidoul:
For now, set 3 variables required by https://github.com/OCA/maintainer-quality-tools/pull/556

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -77,6 +77,7 @@ setuptools.setup(
             'oca-update-pre-commit-excluded-addons = '
             'tools.update_pre_commit_excluded_addons:main',
             'oca-fix-manifest-website = tools.fix_manifest_website:main',
+            'oca-configure-travis= tools.configure_travis:main',
         ],
     },
 )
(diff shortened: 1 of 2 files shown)


sbidoul:
@pedrobaeza have you seen this one? Don't forget to run it on new repos.

About travis-ci.com, I suggest we stay on travis-ci.org until travis migrates everything at once. Otherwise we'll have additional difficulties to maintain both.

pedrobaeza:
No, I didn't, but this is for travis-ci.org, and I don't want that new repos stay on the old infrastructure, which is more complicated to setup and it's already deprecated. Please update for the new infrastructure.

sbidoul:
Problem is I have no way to know which repo is on which travis.

In the meantime, don't forget to add the `GITHUB_*` variables in the new repos.

sbidoul:
I think we must stay on one travis infrastructure until travis has a proper way to migrate all projects at once, and they are not ready yet.

pedrobaeza:
Is this valid for travis-ci.com?

sbidoul:
I'm afraid it is not. :crying_cat_face:

pedrobaeza:
Oh, what a pity. Well, if you want, we can split the work of changing the token. Please mail me with the details and I'll do for a bunch of them.

PR #353, opened 2018-06-19
Add 'Translate me' badge in readme

sbidoul:
image

Similar to the `runbot | Try me` badge.

### DIFF
--- a/tools/gen_addon_readme.py
+++ b/tools/gen_addon_readme.py
@@ -77,6 +77,17 @@ def make_runbot_badge(runbot_id, branch):
     )
 
 
+def make_weblate_badge(repo_name, branch, addon_name):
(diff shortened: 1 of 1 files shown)


sbidoul:
> Should we added to README example too?

No. This will be added automatically.

pedrobaeza:
Well, I wanted to keep it a bit more for having in the same commit this update and the one in the new issue link, but too late..

sbidoul:
@pedrobaeza for now it requires a manual deployment on the OCA server to become active, so no harm done.

pedrobaeza:
Ah, right! Then the manual README generation I made tonight on 2 modules on PRs will be rolled-back, hehe. I'll tell you about the issue link

PR #360, opened 2018-07-23
[IMP] read runbot ids online

sbidoul:
This is to avoid redeploying maintainer-tools on the oca server for the bot each time a new repo is added.


sbidoul:
No objections with this one?

PR #363, opened 2018-07-29
Improve python 3 compatibility

sbidoul:
This makes oca-main-branch-bot python 3 compatible while still being able to run on python 2 branches.

This is necessary to support some 11.0 manifest that have non ascii characters in name or authors, and can be parsed by python 3 and not python2.

### DIFF
--- a/tools/main_branch_bot.py
+++ b/tools/main_branch_bot.py
@@ -14,6 +14,16 @@ from .oca_projects import (
 from .dist_to_simple_index import dist_to_simple_index
 
 
+def _get_python(branch, python2):
+    version = tuple(map(int, branch.split('.')))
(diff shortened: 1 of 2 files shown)


tools/main_branch_bot.py
```diff
@@ -14,6 +14,16 @@
 from .dist_to_simple_index import dist_to_simple_index
 
 
+def _get_python(branch, python2):
+    version = tuple(map(int, branch.split('.')))
+    if version < (11, 0):
+        return python2
```
pedrobaeza: Shouldn't this be text?

pedrobaeza: Or the contrary on the other part?

PR #368, opened 2018-08-20
Relax setuptools version constraint.

sbidoul:
Requiring setuptools<31 now breaks travis builds. This was bound to happen
sooner or later. Allowing the latest setuptools version
has only one drawback: it obliges people using wheels built
with recent versions to always use odoo-autodiscover>=2 on Odoo<11.
This is the safest and easiest anyway and it is best practice
so most people should be unaffected.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -38,11 +38,6 @@ setuptools.setup(
         'setuptools-odoo>=2.1',
         'twine',
         'wheel',
-        # We keep setuptools<31 as long as we can, so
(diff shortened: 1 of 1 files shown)


sbidoul:
> Welcome back to work!

Thanks @pedrobaeza (2 weeks without checking work email, achievement unlocked :).

elicoidal:
@sbidoul welcome back!
(really no mail in 15 days????)

PR #371, opened 2018-09-17
Remove obsolete repos

elicoidal:


elicoidal:
@pedrobaeza do you have the new name so that I can proceed.
* `bank-statement-reconcile` -> `account-reconcile`
* `banking` -> `bank-payment`?
I do not know why but bank-statement-reconcile is still present in the runbot (but not banking)

elicoidal:
OK: I will proceed and cleanse the runbot too

elicoidal:
@pedrobaeza

elicoidal:
Changed the runbot for account-reconcile (nothing to do for banking)
screenshot from 2018-09-26 11-49-02

PR #372, opened 2018-09-17
[FIX] avoid error on creating runbot budget if repo runbot still does…

cubells:
With these changes, readme can be created without error.

### DIFF
--- a/tools/gen_addon_readme.py
+++ b/tools/gen_addon_readme.py
@@ -141,7 +141,7 @@ def gen_one_addon_readme(repo_name, branch, addon_name, addon_dir, manifest):
                     if fragment[-1] != '\n':
                         fragment += '\n'
                     fragments[fragment_name] = fragment
-    runbot_id = get_runbot_ids()[repo_name]
+    runbot_id = get_runbot_ids().get(repo_name)
     badges = []
     development_status = manifest.get('development_status', 'Beta')
     if development_status in DEVELOPMENT_STATUS_BADGES:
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Well, none of the runbot ids should be missing, and the patch should be to add that repo ID. Which one are you talking about?

@sbidoul do you prefer to silently pass this?

cubells:
I detect the problem in two repos: `connector-telephony` and `oca-custom`.

pedrobaeza:
Please propose the patch then on https://github.com/OCA/maintainer-tools/blob/master/tools/repos_with_ids.txt

tools/gen_addon_readme.py
```diff
@@ -141,7 +141,10 @@ def gen_one_addon_readme(repo_name, branch, addon_name, addon_dir, manifest):
                     if fragment[-1] != '\n':
                         fragment += '\n'
                     fragments[fragment_name] = fragment
-    runbot_id = get_runbot_ids()[repo_name]
```
sbidoul: shorter: `runbot_id = get_runbot_ids()[repo_name].get(repo_name)`

sbidoul:
> do you prefer to silently pass this?

We could support repos without runbot why not, it's a small step towards making this script more generic.

OTOH, it's nice to have a way to detect a missing runbot config.

pedrobaeza:
@cubells can you the changes asked by @sbidoul ?

cubells:
@pedrobaeza @sbidoul done.

cubells:
@pedrobaeza 
A warning is not showed if runbot_id doesn't exist. The reason to do that pr is that user could create readme.rst file **even** if the runbot id doesn't exist: the final readme file will not have the runbot ribbon.
But I can add a warning if you prefer.

pedrobaeza:
Yes, please, see second paragraph of https://github.com/OCA/maintainer-tools/pull/372#issuecomment-421969401

cubells:
@pedrobaeza done in https://github.com/OCA/maintainer-tools/pull/372/files#diff-b5e93c0518459db660ec5afe6d44ad1eR157

PR #374, opened 2018-09-22
add "ready to merge" label

sbidoul:

### DIFF
--- a/tools/set_repo_labels.py
+++ b/tools/set_repo_labels.py
@@ -28,6 +28,7 @@ all_labels = {
     'work in progress': '0052cc',
     'wontfix': 'ffffff',
     'migration': 'd4c5f9',
+    'ready to merge': 'bfdadc',
 }
 
 


tools/set_repo_labels.py
```diff
@@ -28,6 +28,7 @@
     'work in progress': '0052cc',
     'wontfix': 'ffffff',
     'migration': 'd4c5f9',
+    'ready to merge': 'bfdadc',
```
elicoidal: What is the difference with `approved`?
Do we have a document with the suggested usage of each label?

sbidoul: @elicoidal approved is fine with me. I had not noticed it.

I actually used ready to merge from #346, as a first PR to https://github.com/OCA/oca-github-bot/pull/1

sbidoul: @elicoidal after further thinking, I think both are useful: approved when all approvals are ok. Then ready to merge when other conditions (like the PR age are met). I updated https://github.com/OCA/oca-github-bot/pull/1 with this concept.

PR #376, opened 2018-09-26
Add several repos without runbot id

cubells:
Related: https://github.com/OCA/maintainer-tools/pull/372

### DIFF
--- a/tools/repos_with_ids.txt
+++ b/tools/repos_with_ids.txt
@@ -125,6 +125,7 @@
 225|github.com/OCA/website-cms
 226|github.com/OCA/edi
 227|github.com/OCA/intrastat
+228|github.com/OCA/connector-telephony
 229|github.com/OCA/interface-github
 230|github.com/OCA/queue
 236|github.com/OCA/l10n-chile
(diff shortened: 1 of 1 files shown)


tools/repos_with_ids.txt
```diff
@@ -159,3 +159,5 @@
 270|github.com/OCA/connector-spscommerce
 271|github.com/OCA/rest-framework
 272|github.com/OCA/l10n-greece
+273|github.com/OCA/oca-custom
+274|github.com/OCA/connector-telephony
```
pedrobaeza: This one should be 228. Where do you take the number?! You have to check it on runbot itself: https://runbot.odoo-community.org/runbot/repo/github-com-oca-connector-telephony-228 (note the final number)

tools/repos_with_ids.txt
```diff
@@ -159,3 +159,5 @@
 270|github.com/OCA/connector-spscommerce
 271|github.com/OCA/rest-framework
 272|github.com/OCA/l10n-greece
+273|github.com/OCA/oca-custom
```
pedrobaeza: This one is not even in runbot. I'm adding it, and indeed it has this ID.

tools/repos_with_ids.txt
cubells: @pedrobaeza done.

PR #384, opened 2018-10-09
update copy maintainers for v11

gurneyalex:
the values of privacy visibility have changed


gurneyalex:
tested live before proposing for merge :smile_cat:

PR #386, opened 2018-10-18
CONTRIBUTED.md found a new home

sbidoul:
- [x] waiting for https://github.com/OCA/odoo-community.org/pull/27


CONTRIBUTING.md
```diff
@@ -1,1199 +1,3 @@
...
-
- * You need to keep the license of the module coded by Odoo SA
- * You need to add the OCA as author (and Odoo SA of course)
- * You need to make the module "OCA compatible" : PEP8, OCA convention and so
-   on so it won't break our CI like runbot, Travis and so.
- * You need to add a disclaimer in the Readme with the following text:
-```
-**This module is a backport from Odoo SA and as such, it is not included in the OCA CLA. That means we do not have a copy of the copyright on it like all other OCA modules.**
-```
+This document has moved to https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst
```
elicoidal: I have a question here: shouldnot we use the canonical URL instead?
If later on we generate with sphinx and static pages it might make sense.
Or we could provide both URL such as:
```
This document has moved to https:odoo-community.org/contributing [not sure about the exact location]
If you wish to contribute, please open a Pull Request to the source document available here: https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst
```

sbidoul: @elicoidal we don't have such a canonical url for this document so far. I could create one but so far it's manual. So I'd prefer we clarify the strategy about our website first. Currently the website links to this github document directly.

dreispt: Can't we use a redirect page on the OCA website, or even an URL shortener service? That would gice us flexibility.

dreispt:
@sbejaoui Thank you for going ahead with this move.

CONTRIBUTING.md
sbidoul: The current redirect mechanism on the OCA website is me maintaining redirections in the nginx config. Until we have something better, I propose we move forward with this PR. The only goal here is to avoid the same document being maintained in 2 different places.

pedrobaeza: @sbidoul AFAIK, you can do the redirections through Odoo itself.

sbidoul: @pedrobaeza I did not find it, do you have a pointer for me?

sbidoul: @pedrobaeza that's a menu entry, so far I knew :)

What I think @dreispt wants is a short, stable url, so we need a kind of url redirector. The was website_seo_redirection but it's not migrated yet.

elicoidal:
It looks like Pedro has a simple way to fix our redirect problem.
I am not blocking this PR because I understand we need to shutdown this doc ASAP so :+1: for me

sbidoul:
Thanks @pedrobaeza that's what I was looking for. I made the redirection and adjusted this PR accordingly.

sbidoul:
Can we merge this. We have the same document at two places and the start diverging...

pedrobaeza:
Ah, right, you mean #389. You have to:

- Fix the conflicts here
- Put the changes in the new file

And then I will merge this.

sbidoul:
Merged manually at c52efcb

PR #389, opened 2018-10-26
[DOC] Require suffix for XML file names

### LINKED ISSUE #388: Discrepancy between doc and module template on how XML files should be named
According to [the documentation](https://github.com/OCA/maintainer-tools/blob/9766529146bcb61ce746623da74a23163199bdd3/CONTRIBUTING.md#user-content-file-naming), XML files should be named without suffix, i.e. data/`res_partner.xml`, `views/res_partner.xml`, etc.

However, if we check the module template, we'll notice  it's with suffix for [data](https://github.com/OCA/maintainer-tools/blob/9766529146bcb61ce746623da74a23163199bdd3/template/module/data/model_name_data.xml), [demo](https://github.com/OCA/maintainer-tools/blob/9766529146bcb61ce746623da74a23163199bdd3/template/module/demo/model_name_demo.xml) and [views](https://github.com/OCA/maintainer-tools/blob/9766529146bcb61ce746623da74a23163199bdd3/template/module/views/model_name_view.xml).

To solve this, I see three possibilities:
1. Modify the documentation to match the template
1. Modify the template to match the documentation
1. Modify the documentation to encourage both alternatives

luisg123v:
Currently, there is a discrepancy between the documentation and the
module template, regarding how XML files should be named. The
documentation says they should be without suffix
(views/res_partner.xml), but the model template says it should be with
suffix (views/res_partner_view.xml).

To be consistent, this modifies the documentation so it matches the
module template.

Closes #388

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -137,22 +137,22 @@ A module is organized in a few directories:
 ### File naming
 
 For `models`, `views` and `data` declarations, split files by the model
-involved, either created or inherited. These files should be named after the
-model. For example, demo data for res.partner should go in a file named
-`demo/res_partner.xml` and a view for partner should go in a file named
-`views/res_partner.xml`. An exception can be made when the model is a
-model intended to be used only as a one2many model nested on the main
-model. In this case, you can include the model definition inside it.
+involved, either created or inherited. When they are XML files, a suffix should
+be included with its category. For example, demo data for res.partner should go
+in a file named `demo/res_partner_demo.xml` and a view for partner should go in
+a file named `views/res_partner_view.xml`. An exception can be made when the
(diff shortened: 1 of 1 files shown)


moylop260:
CI is not related.
Merging since that is a old pending finished

elicoidal:
@moylop260 sorry for the inconvenience but you will need to repeat the PR to the new location of the contributing.md which is https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst

luisg123v:
@elicoidal  thanks for pointing out, here is the PR: https://github.com/OCA/odoo-community.org/pull/29

Shouldn't the file of this project be removed?

sbidoul:
Done at #386

PR #395, opened 2018-11-16
Most of the main branch bot has moved to oca-github-bot

sbidoul:


sbidoul:
@lmignon absolutely. They also [run at night using celery beat,](https://github.com/OCA/oca-github-bot/blob/959b02d3cda6594e26808002885cb2381d5732b8/src/oca_github_bot/cron.py#L13-L17) in case the webhook did not catch it for some reason.

Additional benefits include:
- full visibility on what's going on to all OCA contributors (nothing hidden in obscure cron jobs)
- error handling using sentry

Still TODO in that area: #12 (cla bot), #13 (github teams sync), #15 (wheel generator), #16 (pypi publisher)

PR #397, opened 2018-11-27
[IMP] oca-gen-addon-readme : make organization name configurable and fix images URL

legalsylvain:
Hi all, 

First, thanks for this great tool.

This PR adds 2 improvements on oca-gen-addon-readme (the second depends on the first).

Commit 1 : make the organization name configurable, to let the possibility to generate readme for incubator / custom repositories. (= non OCA repo)

for that purpose, 
-> add a new argument ``--org-name`` with default value set to ``OCA``
-> replace hard coded 'OCA' by the org_name, in all the URLs
-> hide some parts of the template, if the organization is not OCA

Commit 2 : Fix a recurring trouble on figures. In a few words, relative path of figures are not working in the same time on github, on odoo, and on odoo appstore. So, add a little script to translate relative path into full github path. (using rawcontent).
For the time being, to make the image available in all places, maintainers has to write absolute url. (sample in mis-builder : https://raw.githubusercontent.com/OCA/mis-builder/10.0/mis_builder/readme/USAGE.rst)
but the problem are : 
- the url depends on version, so should be changed in each new release ;
- doesnt work on github, when the PR is not merged ; (harder to review)
- longer url ;

sample of the generated code, for a non OCA module : 
https://github.com/grap/grap-odoo-incubator/tree/8.0/account_invoice_supplierinfo_update_triple_discount

(DESCRIPTION.rst contains relative path, and generated readme contains absolute path)

thanks for your review. CC : @sbidoul

### DIFF
--- a/template/module/readme/CONFIGURE.rst
+++ b/template/module/readme/CONFIGURE.rst
@@ -5,6 +5,6 @@ To configure this module, you need to:
 
 #. Go to ...
 
-.. figure:: path/to/local/image.png
+.. figure:: ../static/description/image.png
    :alt: alternative description
    :width: 600 px
--- a/tools/gen_addon_readme.py
+++ b/tools/gen_addon_readme.py
@@ -5,16 +5,23 @@
 import io
 import os
 import re
+import sys
 
 import click
 from docutils import ApplicationError
 from docutils.core import publish_file
 from jinja2 import Template
-from urllib.parse import urljoin
 
 from .gitutils import commit_if_needed
 from .manifest import read_manifest, find_addons, NoManifestFound
 from .runbot_ids import get_runbot_ids
+
+if sys.version_info[0] < 3:
+    # python 2 import
+    from urlparse import urljoin
+else:
+    # python 3 import
+    from urllib.parse import urljoin
 
 
 FRAGMENTS_DIR = 'readme'
--- a/tools/gen_addon_readme.py
+++ b/tools/gen_addon_readme.py
@@ -266,8 +266,10 @@ def gen_one_addon_index(readme_filename):
                    "generated for all installable addons found there.")
 @click.option('--commit/--no-commit',
               help="git commit changes to README.rst, if any.")
+@click.option('--gen-html/--no-gen-html', default=True,
+              help="Generate index html file.")
 def gen_addon_readme(
-        org_name, repo_name, branch, addon_dirs, addons_dir, commit):
+        org_name, repo_name, branch, addon_dirs, addons_dir, commit, gen_html):
     """ Generate README.rst from fragments.
 
     Do nothing if readme/DESCRIPTION.rst is absent, otherwise overwrite
@@ -293,9 +295,10 @@ def gen_addon_readme(
         readme_filename = gen_one_addon_readme(
             org_name, repo_name, branch, addon_name, addon_dir, manifest)
         readme_filenames.append(readme_filename)
-        index_filename = gen_one_addon_index(readme_filename)
-        if index_filename:
-            readme_filenames.append(index_filename)
+        if gen_html:
+            index_filename = gen_one_addon_index(readme_filename)
+            if index_filename:
+                readme_filenames.append(index_filename)
     if commit:
         commit_if_needed(readme_filenames, '[UPD] README.rst')
 
--- a/tools/gen_addon_readme.py
+++ b/tools/gen_addon_readme.py
@@ -138,9 +138,12 @@ def generate_fragment(org_name, repo_name, branch, addon_name, file):
     module_url = "https://raw.githubusercontent.com/{org_name}/{repo_name}"\
         "/{branch}/{addon_name}/".format(**locals())
     for index, fragment_line in enumerate(fragment_lines):
-        if fragment_line.startswith('.. figure::'):
-            path =\
-                fragment_line.replace('\n', '').replace('.. figure:: ', '')
+        if ('.. figure::' in fragment_line or
+                '.. image::' in fragment_line):
+            path = fragment_line.strip()\
+                .replace('\n', '')\
+                .replace('.. figure:: ', '')\
+                .replace('.. image:: ', '')
             if path.startswith('http'):
                 # It is already an absolute path
                 continue
--- a/tools/gen_addon_readme.py
+++ b/tools/gen_addon_readme.py
@@ -1,5 +1,7 @@
 # License AGPLv3 (http://www.gnu.org/licenses/agpl-3.0-standalone.html)
 # Copyright (c) 2018 ACSONE SA/NV
+# Copyright (c) 2018 GRAP (http://www.grap.coop)
+
 import io
 import os
 import re
--- a/tools/gen_addon_readme.py
+++ b/tools/gen_addon_readme.py
@@ -7,6 +7,7 @@ import click
 from docutils import ApplicationError
 from docutils.core import publish_file
 from jinja2 import Template
+from urlparse import urljoin
 
 from .gitutils import commit_if_needed
 from .manifest import read_manifest, find_addons, NoManifestFound
@@ -128,6 +129,35 @@ def make_repo_badge(org_name, repo_name, branch, addon_name):
     )
 
 
+def generate_fragment(org_name, repo_name, branch, addon_name, file):
+    fragment_lines = file.readlines()
+    if not fragment_lines:
+        return False
+
+    # Replace relative path by absolute path for figures
+    module_url = "https://raw.githubusercontent.com/{org_name}/{repo_name}"\
+        "/{branch}/{addon_name}/".format(**locals())
+    for index, fragment_line in enumerate(fragment_lines):
+        if fragment_line.startswith('.. figure::'):
+            path =\
+                fragment_line.replace('\n', '').replace('.. figure:: ', '')
+            if path.startswith('http'):
+                # It is already an absolute path
+                continue
+            else:
+                # remove '../' if exists that make the fragment working
+                # on github interface, in the 'readme' subfolder
+                relative_path = path.replace('../', '')
+                fragment_lines[index] = fragment_line.replace(
+                    path, urljoin(module_url, relative_path))
+    fragment = ''.join(fragment_lines)
+
+    # ensure that there is a new empty line at the end of the fragment
+    if fragment[-1] != '\n':
+        fragment += '\n'
+    return fragment
+
+
 def gen_one_addon_readme(
         org_name, repo_name, branch, addon_name, addon_dir, manifest):
     fragments = {}
@@ -137,10 +167,9 @@ def gen_one_addon_readme(
         )
         if os.path.exists(fragment_filename):
             with io.open(fragment_filename, 'rU', encoding='utf8') as f:
-                fragment = f.read()
+                fragment = generate_fragment(
+                    org_name, repo_name, branch, addon_name, f)
                 if fragment:
-                    if fragment[-1] != '\n':
-                        fragment += '\n'
                     fragments[fragment_name] = fragment
     runbot_id = get_runbot_ids().get(repo_name)
     badges = []
--- a/tools/gen_addon_readme.py
+++ b/tools/gen_addon_readme.py
@@ -117,18 +117,19 @@ def make_weblate_badge(repo_name, branch, addon_name):
     )
 
 
-def make_repo_badge(repo_name, branch, addon_name):
+def make_repo_badge(org_name, repo_name, branch, addon_name):
     badge_repo_name = repo_name.replace('-', '--')
     return (
-        'https://img.shields.io/badge/github-OCA%2F{badge_repo_name}'
+        'https://img.shields.io/badge/github-{org_name}%2F{badge_repo_name}'
         '-lightgray.png?logo=github'.format(**locals()),
-        'https://github.com/OCA/{repo_name}/tree/'
+        'https://github.com/{org_name}/{repo_name}/tree/'
         '{branch}/{addon_name}'.format(**locals()),
-        'OCA/{repo_name}'.format(**locals()),
+        '{org_name}/{repo_name}'.format(**locals()),
     )
 
 
-def gen_one_addon_readme(repo_name, branch, addon_name, addon_dir, manifest):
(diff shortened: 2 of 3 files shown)


tools/gen_addon_readme.py
```diff
@@ -117,29 +118,58 @@ def make_weblate_badge(repo_name, branch, addon_name):
...
+            if path.startswith('http'):
+                # It is already an absolute path
+                continue
+            else:
+                # remove '../' if exists that make the fragment working
+                # on github interface, in the 'readme' subfolder
+                relative_path = path.replace('../', '')
+                fragment_lines[index] = fragment_line.replace(
+                    path, urljoin(module_url, relative_path))
+    fragment = ''.join(fragment_lines)
```
pedrobaeza: Shouldn't you join them with `\n`?

legalsylvain: I was thinking that, but no need, when you use readline, each item ends with '\n'

pedrobaeza: Oh, great

template/module/readme/CONFIGURE.rst
```diff
@@ -5,6 +5,6 @@ To configure this module, you need to:
 
 #. Go to ...
 
-.. figure:: path/to/local/image.png
+.. figure:: ../static/description/image.png
```
pedrobaeza: Maybe this should be ../static/images for being better classified.

legalsylvain: For me, the objective of this commit is to give an exemple with correct relative path (including ../)

I abstain on that point, feel free to make the change directly on my PR if you think it's relevant.

Just to know, is it a new OCA conventions ?

FYI : 
``find ./ -name 'description' -type d -printf '.' | wc -c`` 

> 3905

``find ./ -name 'images' -type d -printf '.' | wc -c``

> 402

(source : All OCA code source, including OCB, OpenUpgrade, ...)

tools/gen_addon_readme.py
```diff
@@ -117,29 +118,58 @@ def make_weblate_badge(repo_name, branch, addon_name):
...
+def generate_fragment(org_name, repo_name, branch, addon_name, file):
+    fragment_lines = file.readlines()
+    if not fragment_lines:
+        return False
+
+    # Replace relative path by absolute path for figures
+    module_url = "https://raw.githubusercontent.com/{org_name}/{repo_name}"\
+        "/{branch}/{addon_name}/".format(**locals())
+    for index, fragment_line in enumerate(fragment_lines):
+        if fragment_line.startswith('.. figure::'):
```
lmignon: @legalsylvain What if the figure is in a indented block? Should we use a regular expression to replace the path what ever its position into the line?

template/module/readme/CONFIGURE.rst
pedrobaeza: No, it's not any guideline, just some random thoughts, as with this we don't need to be in that folder. @sbidoul what do you think?

tools/gen_addon_readme.py
elicoidal: I think we should not only search for `figure` but also for `image` as both are quite common
@legalsylvain

sbidoul:
@legalsylvain thanks for this! I'll review a bit later.

In the meantime make sure the output is 100% identical as before when org is OCA, to avoid regenerating every single readme because a whitespace or line would have been added/removed somewhere.

legalsylvain:
> @sbidoul : In the meantime make sure the output is 100% identical as before when org is OCA, to avoid regenerating every single readme because a whitespace or line would have been added/removed somewhere.

Good idea, I'll check that. In the meantime, when accepting this PR, mass regeneration is expected (for all module with images)

> @elicoidal : I think we should not only search for figure but also for image as both are quite common

indeed, even if I don't understand the interest to have two keywords in the rst syntax. I find that in the OCA both are used indifferently. Don't you think ?

> @lmignon : Should we use a regular expression to replace the path what ever its position into the line?

Excellent idea, indeed. I'll take a look, but any advices are welcome. [here](http://www.commitstrip.com/en/2016/04/08/fing-patterns/) is the summary of my knowledge with regex. So I think [I need help](http://www.commitstrip.com/en/2013/03/06/quelquun-sy-connait-en-expressions-regulieres/) on that point.

sbidoul:
@legalsylvain regeneration for readmes with images is acceptable. But we must be extra careful to reduce noise as much as we can. Thousands of commits mean thousands of travis jobs and thousands of packages pushed to pypi.

legalsylvain:
* [X] Handle ``ìmage`` and ``figure`` (@elicoidal)
* [X] handle indentation (@lmignon)
* [X] Do not generate useless noices regarding ``if org_name == 'OCA'`` (@sbidoul) (tryed on mis-builder/11.0 and server-tools/11.0)

**Note :**
* I did'nt succeed using regex. (sorry, I already lost time on jinja2 syntax :-) ). Feel free to improve the script, commiting directly in the PR, or in another one.
* Making test on OCA / server-tools / 11.0 branch, it's OK for all modules, except one ``module_auto_update``, regarding special char in [author](https://github.com/OCA/server-tools/blob/11.0/module_auto_update/__manifest__.py#L11). As I did'nt change this part (did'nt touch to manifest analysis), I think it's unrelated, but maybe an external point of view could be interesting.

> print authors
``['LasLabs', 'Juan Jos\xc3\xa9 Scaraf\xc3\xada', 'Tecnativa', 'ACSONE SA/NV']``

error in render : 
``UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 8: ordinal not in range(128)``

regards.

sbidoul:
My bet is the unicode error will not appear if you run the script under python3.

pedrobaeza:
Yes, v11 manifest should be parsed with Python 3, not Python 2.

legalsylvain:
OK, thanks for the precisions. so unrelated with that PR.
Is it now all OK for you ?
thanks for all the reviewers.

tools/gen_addon_readme.py
```diff
@@ -117,29 +118,61 @@ def make_weblate_badge(repo_name, branch, addon_name):
...
+            else:
+                # remove '../' if exists that make the fragment working
+                # on github interface, in the 'readme' subfolder
+                relative_path = path.replace('../', '')
+                fragment_lines[index] = fragment_line.replace(
+                    path, urljoin(module_url, relative_path))
+    fragment = ''.join(fragment_lines)
+
+    # ensure that there is a new empty line at the end of the fragment
+    if fragment[-1] != '\n':
```
sbidoul: if fragment is an empty string, this will break

tools/gen_addon_readme.py
```diff
@@ -117,29 +118,61 @@ def make_weblate_badge(repo_name, branch, addon_name):
...
+    module_url = "https://raw.githubusercontent.com/{org_name}/{repo_name}"\
+        "/{branch}/{addon_name}/".format(**locals())
+    for index, fragment_line in enumerate(fragment_lines):
+        if ('.. figure::' in fragment_line or
+                '.. image::' in fragment_line):
+            path = fragment_line.strip()\
+                .replace('\n', '')\
+                .replace('.. figure:: ', '')\
+                .replace('.. image:: ', '')
+            if path.startswith('http'):
```
sbidoul:
```python
image_path_re = re.compile(r'\s*\.\. (figure|image)::\s+(?P<path>.*?)\s*$')
...
mo = image_path_re.match(fragment_line)
if not mo:
    continue
path = mo.group('path')
```

tools/gen_addon_readme.py
```diff
@@ -117,29 +118,61 @@ def make_weblate_badge(repo_name, branch, addon_name):
...
         if os.path.exists(fragment_filename):
             with io.open(fragment_filename, 'rU', encoding='utf8') as f:
-                fragment = f.read()
+                fragment = generate_fragment(
+                    org_name, repo_name, branch, addon_name, f)
                 if fragment:
-                    if fragment[-1] != '\n':
-                        fragment += '\n'
                     fragments[fragment_name] = fragment
     runbot_id = get_runbot_ids().get(repo_name)
```
sbidoul: I suggest looking up runbot_id only if org is OCA?

tools/gen_addon_readme.py
pedrobaeza: You can use `fragment[-1:]` as safer expression that gets the same.

legalsylvain: > if fragment is an empty string, this will break

No because the function return before here.

https://github.com/OCA/maintainer-tools/pull/397/files/ced32a4569f71c121ed5b0e67d9c86d8c0c2f9aa#diff-b5e93c0518459db660ec5afe6d44ad1eR134

tools/gen_addon_readme.template
```diff
@@ -64,6 +64,7 @@ Authors
 Maintainers
 ~~~~~~~~~~~
 
+{% if org_name == 'OCA' -%}
```
elicoidal: nice to have: if not OCA, use an alternative description passed as parameter or better define the footer (OCA or other) in external file so that generation is generic

tools/gen_addon_readme.py
legalsylvain: Thanks a lot ! 

done.

tools/gen_addon_readme.py
legalsylvain: indeed, done.

tools/gen_addon_readme.template
legalsylvain: indeed, but not in the scope of this PR. Feel free to open an issue for such improvements.

legalsylvain:
@sbidoul : is it now all ok for you ? thanks for your review.

sbidoul:
@legalsylvain can you check my last two comments? The one about python3 portability is important.

tools/gen_addon_readme.py
```diff
@@ -2,11 +2,13 @@
 # Copyright (c) 2018 ACSONE SA/NV
```
elicoidal: @legalsylvain maybe you should add your name here

tools/gen_addon_readme.py
```diff
@@ -135,24 +136,24 @@ def generate_fragment(org_name, repo_name, branch, addon_name, file):
...
+            continue
+        path = mo.group('path')
+
+        if path.startswith('http'):
+            # It is already an absolute path
+            continue
+        else:
+            # remove '../' if exists that make the fragment working
+            # on github interface, in the 'readme' subfolder
+            relative_path = path.replace('../', '')
```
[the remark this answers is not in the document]
legalsylvain: Well the usual cases in readme folder are  : 
1. **``static/description/my_image.png``** : 
- (before this PR) github : KO ; gen_addon_readme : OK
- (after this PR) github : OK ; gen_addon_readme : OK

2. **``./static/description/my_image.png``** : 
- (before this PR) github : KO ; gen_addon_readme : OK
- (after this PR) github : OK ; gen_addon_readme : OK

3. **``../static/description/my_image.png``** : 
- (before this PR) github : OK (but doesn't exist) ; gen_addon_readme : KO
- (after this PR) github : OK ; gen_addon_readme : OK

i think we should go ahead, and improve this algorithm (or fix the readme files), if the algorithm doesn't match all cases.

legalsylvain:
OK. sorry, I didn't understand. (I'm not very good in python 3, sorry.)

- @sbidoul : I fixed the import library to make it working with python 3. (tested on server-tools 11.0, the unicode bug does'nt happen anymore)

- @elicoidal : added GRAP mention in the header

- @all : I added a simple new option ``--gen-html/--no-gen-html`` (default is True so no changes), to generate or no, html index files.  (Generating readme for [GRAP incubator repository](https://github.com/grap/grap-odoo-incubator), I saw that html index files makes my repository twice as big, without real interest for me)

```
-------------------------------------------------------------------------------
Language                     files          blank        comment           code
-------------------------------------------------------------------------------
HTML                            40           3440            160          13788
Python                         213           1003           1110           4871
XML                            153            513            804           2991
PO File                         54            592           1614           1872
JavaScript                      11            126            170            477
CSS                              4             22             27             92
YAML                             1              8              0             29
Markdown                         1             10              0             24
-------------------------------------------------------------------------------
SUM:                           477           5714           3885          24144
-------------------------------------------------------------------------------
```

Waiting for other reviews.
regards.

sbidoul:
Thanks!

FYI the html index generation was there because it was told somewhere apps.odoo.com provides better visibility to addons with an html description.

elicoidal:
Good job @legalsylvain :smile:

pedrobaeza:
I have also noticed about the HTML is bloating size. Maybe now that we have our apps store, we should disable the HTML generation?

sbidoul:
@legalsylvain both approaches are fine with me.

legalsylvain:
Test on sys.version_info[0] included.
Tested with python 2 and python 3 and it's working.

pedrobaeza:
Yes, ratings criteria are very obscure IMO, so should we focus in our apps store?

elicoidal:
We should focus in our apps store but cannot ignore the official one where for the moment most people go.
All the more that before it is actually nice enough: we need to add images in all modules and improve a lot of README description.

I still would keep versatility whenever possible: being in the 2 stores increases our visibility and SEO

pedrobaeza:
Yes, but HTML generation is generating a lot of noise in our repos with no apparent benefit in Oddo apps store, except that warning, so I vote for disabling HTML generation.

elicoidal:
Today without HTML desc the odoo apps store is rating the module badly so I vote to keep it

pedrobaeza:
And does it matter? We are not going to be shown highlighted any way.

legalsylvain:
Let's go talking on that point (html / or not) in that issue. https://github.com/OCA/maintainer-tools/issues/399

kind regards.

legalsylvain:
5 :+1:.

Do you think extra reviews are required, or could we merge this PR ? 
regards.

sbidoul:
Deployed on the bot.

legalsylvain:
Thanks @sbidoul ! take me in touch if you observe any problem with this version.
regards.

sbidoul:
@legalsylvain we'll see tomorrow morning... cross fingers. We really need to add tests for this, script.

PR #400, opened 2018-12-01
gen_addon_readme tests

sbidoul:


yucer:
Thank you for this work.

It seems like a template for automatic generation of module description.

It would be good to add some little text of how it can be used.

Or is it used from another script?

sbidoul:
@yucer does this answer your question: https://github.com/OCA/maintainer-tools#readme-generator?

PR #403, opened 2019-02-19
Add Apps store link if OCA

elicoidal:
cc @OCA/board

### DIFF
--- a/tools/gen_addon_readme.template
+++ b/tools/gen_addon_readme.template
@@ -95,7 +95,9 @@ Current maintainer{% if manifest.maintainers|length > 1 %}s{% endif %}:
 This module is part of the `{{ org_name }}/{{ repo_name }} <https://github.com/{{ org_name }}/{{ repo_name }}/tree/{{ branch }}/{{ addon_name }}>`_ project on GitHub.
 
 {% if org_name == 'OCA' -%}
-You are welcome to contribute. To learn how please visit https://odoo-community.org/page/Contribute.
+Discover all OCA modules in the `Official OCA Apps store <https://odoo-community.org/shop>`_
+
+You are welcome to contribute. To learn how to, please visit https://odoo-community.org/page/Contribute.
 
 {%- else -%}
 You are welcome to contribute.


pedrobaeza:
For reducing the diff in all READMEs, switch the order of the sentences. You have to check tests also (there are tests for README generation).

sbidoul:
Changes to the readme generator must be considered very carefully, because the nightly regeneration job will trigger many commits and republish all modules to pypi.

elicoidal:
no clear why travis is failing

elicoidal:
@pedrobaeza I understand the diff problem but I prefer this order which makes more sense to me.
@sbidoul Fully agree and I am open to suggestion: this is a follow up about some discussion with the board about increasing the SEO of the shop. 
Maybe we should queue the modifications and wait for another significant one.

pedrobaeza:
@elicoidal they are unconnected things, so the order is not so important, so for me, they should go the other way for decreasing the diff.

elicoidal:
@pedrobaeza the logic is "Here is the work we do hence if you want to contribute ..."
The other way around is "Contribute. Here is the work we do"

This is why I propose this order

pedrobaeza:
Well, I still think the other way is correct also, so I prefer to not switch it.

elicoidal:
I am not convinced but have no time to fight this

pedrobaeza:
Please rebase to get a green Travis

elicoidal:
@pedrobaeza I dont know how to interpret the travis and fix the PR. Can you help me?

pedrobaeza:
You have to change tests in https://github.com/OCA/maintainer-tools/blob/master/tests/test_gen_addon_readme.py

elicoidal:
> You have to change test

@pedrobaeza Sorry I do not know how to do that... Help is welcome to finish this.

pedrobaeza:
What do you mean by "Our app store is not in very good shape". I use as my first OCA modules info source with no problems.

dreispt (approved):

PR #406, opened 2019-02-27
[IMP] do not fail if setup directory is absent

sbidoul:
Useful for empty repos.

### DIFF
--- a/tools/main_branch_bot.py
+++ b/tools/main_branch_bot.py
@@ -59,16 +59,17 @@ def main(target, repo, branch, push, python2):
             try:
                 # make dists and wheels for each installable addon
                 # and _metapackage
-                sys.stderr.write(
-                    "============> dist_to_simple_index in %s@%s\n" %
-                    (repo, branch),
-                )
-                setup_dirs = [opj('setup', d) for d in os.listdir('setup')]
-                dist_to_simple_index(
-                    target, setup_dirs,
-                    # use the right python to generate wheels
-                    python=_get_python(branch, python2),
-                )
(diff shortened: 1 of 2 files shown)


tools/main_branch_bot.py
```diff
@@ -59,16 +59,17 @@ def main(target, repo, branch, push, python2):
...
-                    "============> dist_to_simple_index in %s@%s\n" %
-                    (repo, branch),
-                )
-                setup_dirs = [opj('setup', d) for d in os.listdir('setup')]
-                dist_to_simple_index(
-                    target, setup_dirs,
-                    # use the right python to generate wheels
-                    python=_get_python(branch, python2),
-                )
+                if os.path.isdir('setup'):
```
pedrobaeza: For reducing the diff and cyclomatic complexity, can we do?
```
if not os.path.isdir('setup'):
    continue
```

sbidoul: I considered that but decided putting a continue there harmed readability (because of the enclosing with and try/finally.

pedrobaeza: Can we extract the condition out of the try?

sbidoul: Yes, not out of the with. Anyway this script is going to die sooner or later (https://github.com/OCA/oca-github-bot/issues/15).

pedrobaeza: OK, no more nitpickings by my part then.

sbidoul:
Travis is green now.

pedrobaeza:
Thanks, merging then.

PR #407, opened 2019-02-28
Rebase

elicoidal:


pedrobaeza:
Why this? There are redundant commits that were already in the repo. Have you mangled the commit history in the repo?

pedrobaeza:
OK, I see this is over another branch, but you shouldn't use main repo for that branches, but on a fork.

elicoidal:
Agree: it was supposed to be a quick fix and I was lazy. I will create a fork next time @pedrobaeza

PR #411, opened 2019-03-21
[ADD] tools: icon generator

mreficent:
### Icon generator

To provide an icon for our modules we generate them automatically.

To generate the icon for the module `auth_keycloak`:

    $ oca-gen-addon-icon --addon-dir=auth_keycloak

----
My idea is to implement it after in the OCA github bot.

### DIFF
--- a/tools/gen_addon_icon.py
+++ b/tools/gen_addon_icon.py
@@ -1,0 +1,80 @@
+# License AGPLv3 (http://www.gnu.org/licenses/agpl-3.0-standalone.html)
+# Copyright (c) 2019 Eficent Business and IT Consulting Services S.L.
+#        (http://www.eficent.com)
+
+import os
+import shutil
+import click
+
+from .gitutils import commit_if_needed
+from .manifest import read_manifest, find_addons, NoManifestFound
+
+
+ICONS_DIR = os.path.join('static', 'description')
+
+ICON_TYPE = 'png'
+
+ICON_TYPES = ['png', 'svg']
+
+
+def gen_one_addon_icon(icon_dir, filetype=ICON_TYPE):
+    icon_filename = os.path.join(icon_dir, 'icon.%s' % filetype)
+    template_filename = \
+        os.path.join(os.path.dirname(__file__).rpartition('tools')[0],
+                     'template', 'module',
+                     ICONS_DIR, 'icon.%s' % filetype)
+    if os.path.exists(template_filename):
+        if not os.path.exists(icon_dir):
+            os.makedirs(icon_dir)
+        shutil.copyfile(template_filename, icon_filename)
+    return icon_filename
+
+
+@click.command()
+@click.option('--addon-dir', 'addon_dirs',
+              type=click.Path(dir_okay=True, file_okay=False, exists=True),
+              multiple=True,
+              help="Directory where addon manifest is located. This option "
+                   "may be repeated.")
+@click.option('--addons-dir',
+              type=click.Path(dir_okay=True, file_okay=False, exists=True),
+              help="Directory containing several addons, the icon will be "
+                   "put for all installable addons found there.")
+@click.option('--commit/--no-commit',
+              help="git commit icon, if not any.")
+def gen_addon_icon(addon_dirs, addons_dir, commit):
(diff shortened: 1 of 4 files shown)


pedrobaeza:
And the icon what contains?

mreficent:
[Default icon](https://github.com/OCA/maintainer-tools/blob/master/template/module/static/description/icon.png)

pedrobaeza:
Then the SVG shouldn't be copied

pedrobaeza:
Any way, I'm not sure about adding this automation. @sbidoul what do you think?

pedrobaeza:
@elicoidal and you, what do you think?

mreficent:
I don't understand this error:
```
stdout = b'Traceback (most recent call last):\n  File "/opt/python/3.6.3/lib/python3.6/runpy.py", line 193, in _run_module_as_m...No such file or directory: \'/tmp/pytest-of-travis/pytest-0/test_gen_addon_icon0/addon/static/description/icon.png\'\n'
```

It's obviously failing in:
```python
with io.open(icon_filename, 'wb') as rf:
```
but I don't know **why**. If the file doesn't exist, the 'w' mode should create it 🤔

elicoidal:
I am not sure either. I see some potential side effects (eg: icon existing in wrong location).
icon adds a bit of visibility in the apps store but the best is nice worked picture or screenshot.
So not against but not sure it has a real utility.

tools/gen_addon_icon.py
```diff
@@ -0,0 +1,96 @@
...
+        # svg was created manually
+        return
+    template_filename = \
+        os.path.join(os.path.dirname(__file__).rpartition('tools')[0],
+                     'template', 'module',
+                     ICONS_DIR, 'icon.svg')
+    with io.open(template_filename, 'rb') as tf:
+        icon = tf.read()
+    with io.open(svg_filename, 'wb') as rf:
+        rf.write(icon)
```
sbidoul: Look at shutil.copyfile instead.

tools/gen_addon_icon.py
```diff
@@ -0,0 +1,96 @@
...
+    return icon_filename
+
+
+def gen_one_addon_svg(icon_filename):
+    icon_dir = os.path.dirname(icon_filename)
+    svg_filename = os.path.join(icon_dir, 'icon.svg')
+    if os.path.exists(svg_filename):
+        # svg was created manually
+        return
```
sbidoul: I you test this here, you should do the same in gen_one_addon_icon too, for symmetry.

mreficent: well, I don't test this in `gen_one_addon_icon` because `gen_one_addon_icon` is only called in `gen_addon_icon` and it's already tested in `gen_addon_icon` before calling `gen_one_addon_icon`.

sbidoul: Yes, that's what I say, it's assymetric and the code path is harder to follow. The could would be much simpler if both function test for the existence of the target filename.

mreficent: ah, ok

pedrobaeza:
As said, you shouldn't include the SVG file.

mreficent:
@pedrobaeza `@click.option('--gen-svg/--no-gen-svg', default=False,` means that by default the svg file will not be included.

pedrobaeza:
But my point is that it shouldn't even been included, as there's no sense of including an unmodified one:

* it's not displayed anywhere
* it takes space and diff
* if you want to make a derivative one, you have the source in one place.

mreficent:
@sbidoul with shutil.copyfile the travis ci is showing the same error as before :S

mreficent:
@sbidoul perhaps the problem is that it cannot create the file in that temporal path? permissions to write?

sbidoul:
I did not really expect shutil.copyfile to change the error... Can you show the full stack trace?

mreficent:
@sbidoul https://github.com/OCA/maintainer-tools/pull/411#issuecomment-475383575

mreficent:
YEEEES, I have found the **why** of the error, I will push the solution in a moment.

mreficent:
Finally!! It's green :green_apple: !! Ready to merge! 🎉🎉🎉

PR #412, opened 2019-03-21
[FIX] tools.gen_addon_readme: deduplicate manifest reading

mreficent:
Problem:
The code it's reading the manifest twice. The second is in the `try` one.
Thus, the first reading of the manifest is useless. Furthermore, you may have an unexpected `NoManifestFound` error in the first reading because the absence of the `try`.

Solution:
Delete the second one and put a `try` in the first.

### DIFF
--- a/tools/gen_addon_readme.py
+++ b/tools/gen_addon_readme.py
@@ -295,14 +295,13 @@ def gen_addon_readme(
         addons.extend(find_addons(addons_dir))
     for addon_dir in addon_dirs:
         addon_name = os.path.basename(os.path.abspath(addon_dir))
-        manifest = read_manifest(addon_dir)
-        addons.append((addon_name, addon_dir, manifest))
-    readme_filenames = []
-    for addon_name, addon_dir, manifest in addons:
         try:
             manifest = read_manifest(addon_dir)
         except NoManifestFound:
             continue
+        addons.append((addon_name, addon_dir, manifest))
+    readme_filenames = []
+    for addon_name, addon_dir, manifest in addons:
         if not os.path.exists(
                 os.path.join(addon_dir, FRAGMENTS_DIR, 'DESCRIPTION.rst')):
             continue


mreficent:
ready :green_apple:

tools/gen_addon_readme.py
```diff
@@ -295,14 +295,13 @@ def gen_addon_readme(
...
     for addon_dir in addon_dirs:
         addon_name = os.path.basename(os.path.abspath(addon_dir))
-        manifest = read_manifest(addon_dir)
-        addons.append((addon_name, addon_dir, manifest))
-    readme_filenames = []
-    for addon_name, addon_dir, manifest in addons:
         try:
             manifest = read_manifest(addon_dir)
         except NoManifestFound:
             continue
```
pedrobaeza: You must put something like `manifest = False` and check that inside the loop for doing the continue

mreficent: But that's something silly. Sooner or later it's filtered. Why filter it later instead of sooner? Furthermore, if done later, you are repeating the manifest reading for the addons introduced with addons_dir (directory) parameter.

pedrobaeza: Well, it's not silly as beginning with, you will have an error after when adding the tuple. If the continue can be skipped inside the for is a different question.

pedrobaeza:
@elicoidal you have merged this without solving the problem I have discovered in https://github.com/OCA/maintainer-tools/pull/412#discussion_r267932767

elicoidal:
@pedrobaeza Sorry: I saw the green approval. Best is to change the approval if something blocking

elicoidal:
Shall I revert or wait for fixing?

pedrobaeza:
Let's way for a fixing. It's not a common situation you find a directory that is not a module.

mreficent:
Sorry, but I don't see the supposed error you are saying in https://github.com/OCA/maintainer-tools/pull/412#discussion_r267932767.

@pedrobaeza As it is now, with the PR merged, please, show me an example with detail where this fails.

As far as I understand, you can have `--addon-dir` or `--addons-dir` or both:

a) If you have `--addons-dir` (`addons_dir` parameter) then this code:
```python
if addons_dir:
    addons.extend(find_addons(addons_dir))
```
adds in `addons` tuples of modules that have a manifest because `find_addons` method filters by manifest.

b) If you have `--addon-dir` (`addon_dirs` parameter) then this code:
```python
for addon_dir in addon_dirs:
    addon_name = os.path.basename(os.path.abspath(addon_dir))
    try:
        manifest = read_manifest(addon_dir)
    except NoManifestFound:
        continue
    addons.append((addon_name, addon_dir, manifest))
```
adds in `addons` tuples of modules that have a manifest because if they don't have a manifest, they will be skipped in the except.

Thus, by a) or by b) or by both, you will have in `addons` a list of tuples of modules with a manifest. And thus, when you do:
```python
for addon_name, addon_dir, manifest in addons:
```
you will always have a manifest, and nothing will break in anyplace. You don't want to create a readme of a module that doesn't have manifest.

pedrobaeza:
@mreficent if the try gives error, `manifest` doesn't get any value, and thus `addons.append((addon_name, addon_dir, manifest))` will give an error, so you have to put on the except part `manifest = False`

mreficent:
@pedrobaeza if the `try` gives error, `manifest` doesn't get any value, and thus in the `except` the `continue ` ends the code and `addons.append((addon_name, addon_dir, manifest))` is not executed. The same way is done in https://github.com/OCA/maintainer-tools/blob/master/tools/manifest.py#L40.

pedrobaeza:
OK, I get you now, as I wasn't counting the outer `for`. OK then. It's correct, and nothing more has to be done.

mreficent:
Good! :+1:

PR #414, opened 2019-03-27
[WIP] maintainers-tools: Readme with local language support

victormmtorres:
### Milestone (Odoo version)
- 

### Module(s)
- 

### Fixes / new features
- Add language localization

cc @Tecnativa

### DIFF
--- a/tools/gen_addon_readme.py
+++ b/tools/gen_addon_readme.py
@@ -172,7 +172,8 @@ def generate_fragment(org_name, repo_name, branch, addon_name, file):
 
 
 def gen_one_addon_readme(
-        org_name, repo_name, branch, addon_name, addon_dir, manifest):
+        org_name, repo_name, branch,
+        addon_name, addon_dir, manifest, code_lang=False):
     fragments = {}
     for fragment_name in FRAGMENTS:
         fragment_filename = os.path.join(
@@ -210,9 +211,14 @@ def gen_one_addon_readme(
         # a real author, but is rather referenced in the
         # maintainers section
     ]
+    if code_lang:
+        template_filename_lang = 'gen_addon_readme.template.{}'.format(
+            code_lang)
+        template_filename = os.path.join(os.path.dirname(__file__), template_filename_lang)
     # generate
-    template_filename = \
-        os.path.join(os.path.dirname(__file__), 'gen_addon_readme.template')
+    if not template_filename:
+        template_filename = \
+            os.path.join(os.path.dirname(__file__), 'gen_addon_readme.template')
     readme_filename = \
         os.path.join(addon_dir, 'README.rst')
     with io.open(template_filename, 'rU', encoding='utf8') as tf:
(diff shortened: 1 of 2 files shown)


tools/gen_addon_readme.template.es
```diff
@@ -0,0 +1,104 @@
...
+{% endfor %}
+{% for _ in badges %}|badge{{ loop.index }}| {% endfor %}
+
+{{ fragments.get('DESCRIPTION', '') }}
+**Table of contents**
+
+.. contents::
+   :local:
+
+{{ fragment('INSTALL', 'Instalción') }}
```
elicoidal:
```suggestion
{{ fragment('INSTALL', 'Instalación') }}
```

tools/gen_addon_readme.template.es
```diff
@@ -0,0 +1,104 @@
...
+{{- fragment('CONFIGURE', 'Configuración') }}
+{{- fragment('USAGE', 'Uso') }}
+{{- fragment('DEVELOP', 'Desarrollo') }}
+{{- fragment('ROADMAP', 'Incidencias conocidas / Hoja de ruta') -}}
+{{- fragment('HISTORY', 'Changelog') -}}
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues <https://github.com/{{ org_name }}/{{ repo_name }}/issues>`_.
```
elicoidal: Why not translate all the content?

tools/gen_addon_readme.py
```diff
@@ -282,8 +289,13 @@ def gen_one_addon_index(readme_filename):
               help="git commit changes to README.rst, if any.")
 @click.option('--gen-html/--no-gen-html', default=True,
               help="Generate index html file.")
+@click.option('--lang', default='',
+              help="Target language file code (default is English)"
+                   "List of supported codes: "
+                   "[(--lang=es for Español),]")
```
elicoidal:
```suggestion
                   "[(--lang=es for Spanish),]")
```

elicoidal: Somehow feels short: you should use the country+lang code.
What about MX,etc...

victormmtorres:
Changes done @elicoidal

tools/gen_addon_readme.py
```diff
@@ -282,8 +292,13 @@ def gen_one_addon_index(readme_filename):
               help="git commit changes to README.rst, if any.")
 @click.option('--gen-html/--no-gen-html', default=True,
               help="Generate index html file.")
+@click.option('--lang', default='',
+              help="Target language file code (default is en_En)"
+                   "List of supported codes: "
+                   "[(--lang=es for es_ES),]")
```
elicoidal: I am not sure that covers my point. How would you code MX Spanish?

victormmtorres: Thought in principle to have just one file per language to not do it more complicated but I understand your concern. So what about now with language + country codes?

elicoidal: Indeed that was my intention originally: use standard language+country code @victormmtorres

victormmtorres: Think it's done, could you update your review?

pedrobaeza:
@sbidoul how do you think we can enable this by repositories for example for having the bot populating Spanish template in OCA/l10n-spain?

tools/gen_addon_readme.template.es_Es
```diff
@@ -0,0 +1,102 @@
+﻿{%- macro fragment(name, title, sub='=') %}
```
sbidoul: In the file name: es_Es => es_ES ?

sbidoul:
Instead of passing a `--lang` option we could maybe take the language from the `readme` directory name. So if `readme/` is present, generate for the default english template. If `readme.es_ES/` is present, generate for es_ES, etc.

pedrobaeza:
Good idea, @sbidoul 

@victormmtorres can you implement this?

sbidoul:
Question: should allow for multiple readme's? (README.rst, README.es_ES.rst)

pedrobaeza:
That's something I have think about, and the visibility in the apps store, in Odoo, in GitHub, but starting with, people is not going to translate too much the READMEs, unless we provide a Weblate like platform for that, so for now, I would go for this solution for l10n-xx repos, and then we can think about mechanism for what you say.

rafaelbn:
@victormmtorres please check what @sbidoul said https://github.com/OCA/maintainer-tools/pull/414#issuecomment-502802644

rafaelbn:
@pedrobaeza maybe other team member can finish this one

pedrobaeza:
The approach should be another, so closing.

PR #416, opened 2019-04-01
[IMP] robustify gen-addon-icon

sbidoul:
@mreficent

### DIFF
--- a/tools/gen_addon_icon.py
+++ b/tools/gen_addon_icon.py
@@ -27,7 +27,8 @@ def gen_one_addon_icon(icon_dir, filetype=ICON_TYPE):
         if not os.path.exists(icon_dir):
             os.makedirs(icon_dir)
         shutil.copyfile(template_filename, icon_filename)
-    return icon_filename
+        return icon_filename
+    return None
 
 
 @click.command()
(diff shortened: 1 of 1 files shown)


tools/gen_addon_icon.py
```diff
@@ -27,7 +27,8 @@ def gen_one_addon_icon(icon_dir, filetype=ICON_TYPE):
         if not os.path.exists(icon_dir):
             os.makedirs(icon_dir)
         shutil.copyfile(template_filename, icon_filename)
-    return icon_filename
+        return icon_filename
+    return None
```
mreficent: But this can never happen :S
In theory, `if os.path.exists(template_filename):` condition should be always True because the icon still exists that [path](https://github.com/OCA/maintainer-tools/blob/master/template/module/static/description/icon.png).

sbidoul: Yes, I know this *should* never happen. That's why I said I do not understand the error...

Yet the script fails when run in the context of the bot.

mreficent: This means that in the context of the bot, `os.path.dirname(__file__)` is not working correctly and or it doesn't access correctly to maintaner-tools repo.  🤔

PR #418, opened 2019-04-17
[ADD] manifest: preloadable flag

mreficent:
Odoo does things if the module has a static folder. See https://github.com/OCA/OCB/blob/11.0/odoo/http.py#L1340.

Modules with `'preloadable': False` should not have a static folder.

### DIFF
--- a/template/module/__manifest__.py
+++ b/template/module/__manifest__.py
@@ -15,6 +15,7 @@
     "license": "AGPL-3",
     "application": False,
     "installable": True,
+    "preloadable": True,
     "pre_init_hook": "pre_init_hook",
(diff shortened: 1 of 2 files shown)


sbidoul:
@mreficent I just deployed this in the github bot. Don't forget to ping me when that needs to happen in the future.

yelizariev:
It's the only place where I found an explanation for the tag. So, let me ask it here:

Why don't wrap monkey patches in post_load hook instead of removing static folder?

MiquelRForgeFlow:
That's a good question for the people who has the modules that use this tag. @hbrunn

PR #419, opened 2019-05-11
[FIX] copy_maintainers: detect GitHub teams

sbidoul:
The project names in the OCA database have been changed to allow for a better presentation on the website.

We now need some heuristics to detect the corresponding GitHub teams.


sbidoul:
@gurneyalex this is deployed, your eagle eye is nevertheless welcome here.

pedrobaeza:
Is this deployed on OCA instance?

sbidoul:
yes, this patch is part of the running version

pedrobaeza:
I think we should merge it then.

MiquelRForgeFlow:
What are we waiting here?

PR #425, opened 2019-08-29
Add a warning in readme for Alpha status

guewen:
The Alpha badge is visible but not enough to get people aware
of the consequences of using Alpha addons.

### DIFF
--- a/tools/gen_addon_readme.template
+++ b/tools/gen_addon_readme.template
@@ -26,6 +26,15 @@
 {% for _ in badges %}|badge{{ loop.index }}| {% endfor %}
 
 {{ fragments.get('DESCRIPTION', '') }}
+{% if development_status == 'Alpha' -%}
+
+.. IMPORTANT::
+   This is an alpha version, the data model and design can change at any time without warning.
+   Only for development or testing purpose, do not use in production.
+   `More details on development status <https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/oca_module_lifecycle_development_status.rst>`_
+
+{% endif -%}
+
 **Table of contents**
 
 .. contents::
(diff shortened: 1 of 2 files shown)


sbidoul:
Sorry I had not seen this one. I agree on the principle, but... any change to the readme will cause new commits for all addons, a travis meltdown and re publishing everything to pypi.

Are we sure this is worth it?

tools/gen_addon_readme.template
```diff
@@ -26,6 +26,15 @@
 {% for _ in badges %}|badge{{ loop.index }}| {% endfor %}
 
 {{ fragments.get('DESCRIPTION', '') }}
+{% if development_status == 'Alpha' -%}
+
+.. IMPORTANT::
+   This is an alpha version, the data model and design can change at any time without warning.
+   Only for development or testing purpose, do not use in production.
+   `More details on development status <https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/oca_module_lifecycle_development_status.rst>`_
```
sbidoul: This must be changed to the canonical URL that this mentioned at the bottom of that document.

pedrobaeza:
@sbidoul this only affects in case you put Alpha state, which must be put manually (default one is Beta).

tools/gen_addon_readme.template
pedrobaeza: But I think the canonical URL doesn't point to the exact anchor

sbidoul:
>this only affects in case you put Alpha state

Hm correct. The PR does not change the generated files for other status. Ok then, sorry for the noise.

I'll push a change for the canonical URL.

sbidoul:
Done at f351e95a0371c9afb016f33e1558c91cbff9efe0

This will be in the next bot deployment.

sbidoul:
deployed

PR #432, opened 2019-10-26
[ADD] oca-towncrier

sbidoul:
One difficulty of maintaining changelog (`readme/HISTORY.rst` in OCA addons) is merge conflicts on PRs that touch it.

[towncrier](https://github.com/hawkowl/towncrier) is a tool that lets contributors write news fragment files (in `readme/newsfragments/` in OCA addons). News fragments have the issue/pr number as name and a suffix (.feature, .bugfix, .doc, .removal) indicating the type of changelog entry.

When run, `towncrier` generates a commit that updates `HISTORY.rst` in a standardized manner and removes the news fragment files.

This maintainer tool script does this with a configuration that works well with `oca-gen-addon-readme`. A typical commit produced by this tool can be seen [here](https://github.com/OCA/mis-builder/commit/2f87e1a3d0bef7a7cdbe219d706af3a0ca958ad0).

My idea is to run it in the bumpversion step of "ocabot merge".

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,2 @@
+[flake8]
+max-line-length = 88
--- a/setup.py
+++ b/setup.py
@@ -38,6 +38,8 @@ setuptools.setup(
         'polib',
         'pygments',
         'requests',
+        'toml>=0.10.0',  # for oca-towncrier
+        'towncrier>=19.2',  # for oca-towncrier
         'twine',
         'wheel',
     ],
@@ -67,6 +69,7 @@ setuptools.setup(
             'oca-pypi-upload = tools.pypi_upload:cli',
             'oca-gen-addon-readme = tools.gen_addon_readme:gen_addon_readme',
             'oca-gen-addon-icon = tools.gen_addon_icon:gen_addon_icon',
+            'oca-towncrier = tools.oca_towncrier:oca_towncrier',
         ],
     },
 )
--- a/tests/test_towncrier.py
+++ b/tests/test_towncrier.py
@@ -1,0 +1,43 @@
+# License AGPLv3 (http://www.gnu.org/licenses/agpl-3.0-standalone.html)
+# Copyright (c) 2018 ACSONE SA/NV
+
+import toml
+
+from tools.oca_towncrier import (
+    _make_issue_format,
+    _preserve_file,
+    _prepare_pyproject_toml,
+)
+
+
+def test_make_issue_format():
+    assert (
+        _make_issue_format("OCA", "repo")
+        == "`#{issue} <https://github.com/OCA/repo/issues/{issue}>`_"
+    )
+
+
+def test_preserve_file(tmp_path):
+    p = tmp_path / "dummy"
+    with _preserve_file(str(p)):
+        # path does not exist
+        p.write_text(u"abc")
+    assert not p.exists()
+    p.write_text(u"abc")
+    with _preserve_file(str(p)):
+        p.write_text(u"xyz")
+    assert p.read_text() == u"abc"
+
+
+def test_prepare_pyproject_toml(tmp_path):
+    with _prepare_pyproject_toml(str(tmp_path), "OCA", "repo"):
+        with open(str(tmp_path / "pyproject.toml")) as f:
+            pyproject = toml.load(f)
+            assert set(pyproject["tool"]["towncrier"].keys()) == {
+                "template",
+                "underlines",
+                "title_format",
+                "issue_format",
+                "directory",
+                "filename",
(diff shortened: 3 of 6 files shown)


.flake8
```diff
@@ -0,0 +1,2 @@
+[flake8]
+max-line-length = 88
```
pedrobaeza: Why adding this?

sbidoul: Its a first step towards making this repo compatible with Black's default options (this change is done in MQT already).

MiquelRForgeFlow: But the file needs to end with an empty line, right?

legalsylvain:
Hi @sbidoul. Looks great ! 
two questions : 

- as a contributor, Do I have to add a file ``154.bugfix`` or only a file ``.bugfix`` ? 
I mean, when I make a new branch, I don't know the number of the PR. I see as a little problem if I have to rename the file, once the PR made. (not a big deal, but well...)

I see a problem for people that are running locally ``oca-gen-addon-readme`` that can occur.
If I understand correctly, 
- if I run locally the script, new fragments will be deleted, and history.rst updated but then, when ocabot will run ``oca-gen-addon-readme`` new fragment will not be there, and conflict can occurs. (or maybe I missed something).
If so, it should be great to add an option, to run towncrier, disabled by default, and enabled when called by the ocabot. Don't you think ? 

Sorry if my questions are not relevant. 
kind regards.

sbidoul:
> rename the file, once the PR made

You need to provide a issue or PR number in the file name indeed. This is a very minor annoyance which I experienced too on other projects using towncrier. Let's say this somewhat encourages creating issues before PRs, which is a good practice anyway ;)

Regarding conflicts I don't think it would be a problem because towncrier will update HISTORY.rst which is itself a fragment to be included in the readme by oca-gen-addon-readme. My ideas is to have both running during ocabot merge. I may misunderstand your point though.

legalsylvain:
thanks for your answer @sbidoul 

Point 1, yes. Well, not a blocking point, because it is very optional in OCA, to update the HISTORY.rst file.

Point 2 : 
> Regarding conflicts I don't think it would be a problem because towncrier will update HISTORY.rst which is itself a fragment to be included in the readme by oca-gen-addon-readme. My ideas is to have both running during ocabot merge. I may misunderstand your point though.

Scenario OK : 

- I do a PR 155 against mis_builder with a file ``readme/newsfragments/155.doc``
- You do a PR 156 against mis_builder with a file ``readme/newsfragments/156.bugfix``
- reviewers do the job, and call ocabot that run oca-gen-addon-readme that includes towncrier that remove the fragments files, and update history.rst 
--> all is ok.

Scenario KO : 

- I do a PR 155 against mis_builder with a file ``readme/newsfragments/155.doc``
- I run locally ``oca-gen-addon-readme`` commit and push 
- You do a PR 156 against mis_builder with a file ``readme/newsfragments/156.bugfix``
- You run locally ``oca-gen-addon-readme`` commit and push 
If I understand correctly, in this step, the fragments files do not exist anymore, and history.rst will be conflicted, when merging 155 then 156 PR. 
(Or did I missed something ?)

sbidoul:
> Scenario KO

I see what you mean. oca-towncrier is a separate tool that is *not* meant to be run locally on PR branches. It will run only when doing a "release", i.e.  "ocabot merge (patch|minor|major)". So there should be no conflict in that scenario.

Of course, the OCA peculiarity is that each "ocabot merge" is a release, so HISTORY entries will generally have only each have one news, unless you group several PR on a release branch. 

But at least it's easier to write standalone newfragments than updating the HISTORY.rst because you don't need to ask yourself about the date or version that is to be written in the HISTORY file because that gets delayed until merging.

legalsylvain:
Ok ! Thanks for the explanation and :+1: for the feature. 

About the first point what about to allow to add files like ``whatever.doc`` or ``branch-name.bugfix`` and add a little script in the ocabot to rename the prefix file with the number of the PR before all the stuff you added ? I see it possible, and make easier to contribute.
(The only constraint I see is to have a unique term to avoid conflict if a gitaggregate command merge a lot of pending PR)
It avoid developpers to rename the files when the PR is done (or if the PR is closed and a new one is created, based on the first branch), etc. 
The OCA reality seems to be : 95% of the PR are done after having created the branch, not based on an issue. 
Don't you think ?
Anyway, thanks for this improvment !

sbidoul:
I improved this a little bit and tested again and fixed a couple of issues.
It is now good for merging and we can progress with it's integration in the bot: https://github.com/OCA/oca-github-bot/pull/106

PR #434, opened 2019-10-30
[IMP] Add pre-commit installation session

Tonow-c2c:
Translation from https://github.com/OCA/l10n-brazil/wiki/Pre-commit-hook

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,3 +171,27 @@ egg.
 You can use the `GITHUB_TOKEN` environment variable to specify the token
 
     $ GITHUB_TOKEN=xxx python -m tools.copy_maintainers
+
+**Install pre-commit**
+
+May be this section should be move to https://github.com/OCA/maintainer-tools/wiki#how-to
+
+To ensure your commits are not made without pre-commit check executing this command,
+you can include it in your computer's git hook:
+
+```
+pip install pre-commit
+git config --global core.hooksPath ~/.git/hooks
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Not sure this should be here, as pre-commit is not needed for running maintainer-toolds.

Tonow-c2c:
@pedrobaeza I agree with you may be we should add new repos for _procedure/documentation_ or a section in https://odoo-community.org/page/documentation where people can propose changes
but a very clear place to share procedures, knowledge and best practices

pedrobaeza:
For me that place is the wiki about migration to version 13.0, that is referenced in each migration issue and that is where the change for black is happening.

Tonow-c2c:
So can you add some help on this wiki https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-13.0 to explain how to install pre-commit and use or a link to https://pre-commit.com/ ?

pedrobaeza:
Well, specific tools deployment shouldn't be there. We can add a link to other place, but installing pre-commit shouldn't be something to be put in a migration wiki. There's already the mentioning of pre-commit needing to be run in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-13.0#howto. Reading these instructions, I see that I didn't need to do all this stuff for making it work and this is something unnecessary complicated IMO. @sbidoul's blog post was enough for me.

PR #436, opened 2019-11-28
Provide pre-commit hooks, compatible with prettier

Yajo:
### Fixes / new features
- oca-gen-addons-table now supports setting `--readme` and `--addons-dir` paths, to support non-oca repos structures.
- It also supports a more helpful `--help`, which also suggests you to add [prettier exlusion tags](https://prettier.io/docs/en/ignore.html#markdown), to avoid double-edition of README.md files.
- pre-commit hooks supplied and documented.

@Tecnativa @sbidoul

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,0 +1,20 @@
+- id: oca-gen-addon-readme
+  name: Generate addons README files from fragments
+  always_run: true
+  entry: oca-gen-addon-readme
+  language: python
+  pass_filenames: false
+
+- id: oca-gen-addons-table
+  name: Generate addons table in README
+  always_run: true
+  entry: oca-gen-addons-table
+  language: python
+  pass_filenames: false
+
+- id: oca-gen-addon-icon
+  name: Generate addons icons
+  always_run: true
+  entry: oca-gen-addon-icon
+  language: python
+  pass_filenames: false
--- a/README.md
+++ b/README.md
@@ -105,7 +105,7 @@ To generate the final README for the module `auth_keycloak`:
 
 The result will be a fully PyPI compliant README.rst in the root of your module.
 
-You may also use this script for your own repositories by specifying this 
+You may also use this script for your own repositories by specifying this
 additional argument `--org-name=myorganisation`
 
 
@@ -171,3 +171,32 @@ egg.
 You can use the `GITHUB_TOKEN` environment variable to specify the token
 
     $ GITHUB_TOKEN=xxx python -m tools.copy_maintainers
+
+## Integration with `pre-commit`
+
+In any addons repo, you can use these pre-commit hooks:
+
+```yaml
+# .pre-commit-config.yaml file
+repos:
+  - repo: https://github.com/OCA/maintainer-tools
+    rev: master # This is just an example; you must use a tag/commit instead!
+    hooks:
+      # Use each script's `--help` to understand the args
+      - id: oca-gen-addon-readme
+        args:
+          - --addons-dir=.
+          - --org-name=OCA
+          - --repo-name=server-tools
+          - --branch=13.0
+
+      # This job could easily produce conflicts when it runs on every commit,
+      # so it's added as a manual job. If you automate it, beware.
+      # See https://pre-commit.com/#confining-hooks-to-run-at-certain-stages
+      - id: oca-gen-addons-table
+        stages: [manual]
+
+      - id: oca-gen-addon-icon
+        args:
+          - --addons-dir=.
+```
--- a/tools/gen_addons_table.py
+++ b/tools/gen_addons_table.py
@@ -2,19 +2,19 @@
 #  -*- coding: utf-8 -*-
 # License AGPLv3 (http://www.gnu.org/licenses/agpl-3.0-standalone.html)
 """
-This script replaces markers in the README.md files
+This script replaces markers in the README.md file
 of an OCA repository with the list of addons present
 in the repository. It preserves the marker so it
 can be run again.
 
-The script must be run from the root of the repository,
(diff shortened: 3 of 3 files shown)


sbidoul:
@Yajo this is very interesting.

I would not recommend a pre-commit hook for the addons table in README.md though, because it will create merge conflicts and the generated table would not have the correct addon version because that one is only known later we we do ocabot merge patch|minor|major.

For README.rst and the icon it should not be an issue, so +1 for that part.

Another useful hook would be running `setuptools-odoo-makedefault` which is another thing the bot does it would be useful to have on PR branches.

Yajo:
> Another useful hook would be running `setuptools-odoo-makedefault` which is another thing the bot does it would be useful to have on PR branches.

Could you give me more details on this? I cannot find it.

sbidoul:
This is how the bot does it:

https://github.com/OCA/oca-github-bot/blob/ca2891200c381d10f2625360f1cf9467ad6f14d0/src/oca_github_bot/tasks/main_branch_bot.py#L50-L58

The tools comes from the `setuptools-odoo` package.

i would do the same in a pre-commit hook, except the metapackage part that risks creating merge conflicts too.

Yajo:
OK, that'd be nice, but then the hooks should be provided by the setuptools-odoo repo itself, and then added in precommit in https://github.com/OCA/maintainer-quality-tools/pull/618 or a similar PR. I think this PR has nothing to do in that sense.

I mean, that's a good idea, but shouldn't block this one, right?

sbidoul:
> but shouldn't block this one, right?

Correct. Please just remove the part related to gen-addons-table.

tools/gen_addons_table.py
```diff
@@ -63,17 +63,15 @@ def replace_in_readme(readme_path, header, rows_available, rows_unported):
...
             render_markdown_table(header, rows_available),
             '\n'
         ])
     if rows_unported:
         addons.extend([
             '\n',
             '\n',
-            'Unported addons\n',
-            '---------------\n',
+            '## Unported addons\n',
```
sbidoul: I'd remove this from this PR. I agree it's better, but this requires synchronization with the bot, so it's better to have that separate so I can focus on deploying that separately.

README.md
```diff
@@ -171,3 +171,26 @@ egg.
...
+    rev: master # Better use a tag/commit!
+    hooks:
+      # Use each script's `--help` to understand the args
+      - id: oca-gen-addon-readme
+        args:
+          - --addons-dir=.
+          - --org-name=OCA
+          - --repo-name=server-tools
+          - --branch=13.0
+      - id: oca-gen-addons-table
```
sbidoul: remove this

README.md
```diff
@@ -171,3 +171,26 @@ egg.
...
+
+## Integration with `pre-commit`
+
+In any addons repo, you can use these pre-commit hooks:
+
+```yaml
+# .pre-commit-config.yaml file
+repos:
+  - repo: https://github.com/OCA/maintainer-tools
+    rev: master # Better use a tag/commit!
```
sbidoul: With pre-commit using tags (or can we use commit hashes?) is a must.

.pre-commit-hooks.yaml
```diff
@@ -0,0 +1,20 @@
...
+  entry: oca-gen-addon-readme
+  language: python
+  pass_filenames: false
+
+- id: oca-gen-addons-table
+  name: Generate addons table in README
+  always_run: true
+  entry: oca-gen-addons-table
+  language: python
+  pass_filenames: false
```
sbidoul: I would remove this part

tools/gen_addons_table.py
Yajo: I don't understand what you mean... Isn't this what the bot itself writes? So, any change here would be done automatically by it, right?

The reasoning for this is to match prettier heading styles.

In any case, I guess you know what you're talking about, so let me just add a TODO note here. As prettier is disabled in these files for now and, if enabled, the ignore tags would make it not touch these lines, it shouldn't be a problem in any case.

README.md
Yajo: This is just an example that tells you what you *can* do.

Instead of removing it, I added `stages: [manual]`, so this trigger doesn't run automatically. I also added a comment explaining that this could produce conflicts easily and should be used with care.

These are docs anyway, it's not like the feature is buggy, it's just that users need to know what it does and the problems it has.

README.md
Yajo: OK, I changed it.

.pre-commit-hooks.yaml
Yajo: Resolved below.

tools/gen_addons_table.py
```diff
@@ -84,19 +86,24 @@ def replace_in_readme(readme_path, header, rows_available, rows_unported):
...
-@click.command()
+@click.command(help=__doc__)
 @click.option('--commit/--no-commit',
               help="git commit changes to README.rst, if any.")
-def gen_addons_table(commit):
-    readme_path = 'README.md'
+@click.option('--readme-path', default="README.md",
+              type=click.Path(dir_okay=False, file_okay=True, exists=True),
+              help="README.md file with addon table markers")
+@click.option('--addons-dir', default="",
```
sbidoul:
```suggestion
@click.option('--addons-dir', default=".",
```

because `os.listdir("")` fails

sbidoul:
Travis' red

Yajo:
:heavy_check_mark: now

sbidoul:
> the hooks should be provided by the setuptools-odoo repo

The setuptools-odoo-make-default pre-commit hook is available in setuptools-odoo 2.5.3.

PR #442, opened 2020-01-31
[ADD] Massive changer

pedrobaeza:
This script helps to make a massive change across a whole organization on contents of a file.

Installation
============

For using this utility, you need to install these dependencies:

* github3.py library for handling Github calls. To install it, use: `sudo pip install github3.py`.

Configuration
=============

You must have a file called oca.cfg on the same folder of the script for storing credentials parameters. You can generate an skeleton config running this script for a first time.

Usage
=====
oca-massive-change [-h] [-p PROJECTS [PROJECTS ...]] [-e EMAIL]
                   [-t TARGET_ORG] [-m COMMIT_MESSAGE]
                   target path source_string target_string

Performs a massive change over a whole organization.

positional arguments:
  target                Target branch
  path                  File path
  source_string         Source string (or regexp)
  target_string         Target string (or regexp)

optional arguments:
  -h, --help            show this help message and exit
  -p PROJECTS [PROJECTS ...], --projects PROJECTS [PROJECTS ...]
                        List of specific projects to migrate
  -e EMAIL, --email EMAIL
                        Provides an email address used to commit on GitHub if
                        the one associated to the GitHub account is not public
  -t TARGET_ORG, --target-org TARGET_ORG
                        By default, the GitHub organization used is OCA. This
                        arg lets you provide an alternative organization
  -m COMMIT_MESSAGE, --message COMMIT_MESSAGE
                        By default, '[UPD] Massive change [skip ci]', but can
                        be provided with an alternative one

This script will perform the following operations for each project:

* Look for the selected file in the target branch.
* If found, modify it searching for the source string and replacing it by the target string.
* Commit the changes with optional commit message.

cc @Tecnativa

### DIFF
--- a/tools/massive_change.py
+++ b/tools/massive_change.py
@@ -1,0 +1,221 @@
+#!/usr/bin/env python
+#  -*- coding: utf-8 -*-
+# License AGPLv3 (http://www.gnu.org/licenses/agpl-3.0-standalone.html)
+"""
+This script helps to make a massive change across a whole organization
+on contents of a file.
+
+Installation
+============
+
+For using this utility, you need to install these dependencies:
+
+* github3.py library for handling Github calls. To install it, use:
+  ``sudo pip install github3.py``.
+
+Configuration
+=============
+
+You must have a file called oca.cfg on the same folder of the script for
+storing credentials parameters. You can generate an skeleton config running
+this script for a first time.
+
+Usage
+=====
+oca-massive-change [-h] [-p PROJECTS [PROJECTS ...]] [-e EMAIL]
+                   [-t TARGET_ORG] [-m COMMIT_MESSAGE]
+                   target path source_string target_string
+
+Performs a massive change over a whole organization.
+
+positional arguments:
+  target                Target branch
+  path                  File path
+  source_string         Source string (or regexp)
+  target_string         Target string (or regexp)
+
+optional arguments:
+  -h, --help            show this help message and exit
+  -p PROJECTS [PROJECTS ...], --projects PROJECTS [PROJECTS ...]
+                        List of specific projects to migrate
+  -e EMAIL, --email EMAIL
+                        Provides an email address used to commit on GitHub if
+                        the one associated to the GitHub account is not public
+  -t TARGET_ORG, --target-org TARGET_ORG
+                        By default, the GitHub organization used is OCA. This
+                        arg lets you provide an alternative organization
+  -m COMMIT_MESSAGE, --message COMMIT_MESSAGE
+                        By default, '[UPD] Massive change [skip ci]', but can
+                        be provided with an alternative one
+
+This script will perform the following operations for each project:
+
+* Look for the selected file in the target branch.
+* If found, modify it searching for the source string and replacing it by the
+  target string.
+* Commit the changes with optional commit message.
+
+Credits
+=======
+
+Contributors
+------------
+
+* Tecnativa - Pedro M. Baeza
+
+Maintainer
+----------
+
+.. image:: https://odoo-community.org/logo.png
+   :alt: Odoo Community Association
+   :target: https://odoo-community.org
+
+This module is maintained by the OCA.
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose
+mission is to support the collaborative development of Odoo features and
+promote its widespread use.
+
+To contribute to this module, please visit https://odoo-community.org.
+"""
+
+from __future__ import print_function
+import argparse
+import re
+from . import github_login
+from . import oca_projects
+from .config import read_config
+from github3.exceptions import NotFoundError
+
+DEFAULT_MESSAGE = "[UPD] Massive change [skip ci]"
+
+
+class MassiveChanger(object):
+    def __init__(self, target, target_org=None, email=None, message=None):
+        # Read config
+        config = read_config()
+        self.gh_token = config.get('GitHub', 'token')
+        # Connect to GitHub
+        self.github = github_login.login()
+        gh_user = self.github.me()
+        if not gh_user.email and not email:
+            raise Exception(
+                'Email required to commit to github. Please provide one on '
+                'the command line or make the one of your github profile '
+                'public.')
+        self.gh_credentials = {'name': gh_user.name or str(gh_user),
+                               'email': gh_user.email or email}
+        self.gh_target_branch = target
+        self.gh_org = target_org or 'OCA'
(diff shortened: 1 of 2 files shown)


pedrobaeza:
This is using GitHub API. There's no concept of "pushing". The similar operation is:

https://github.com/OCA/maintainer-tools/pull/442/files#diff-ff7702c786dba119e1cda5eaa8acd52cR152

yelizariev:
I'd suggest name **OCA sed** tool or something like that. 
Does OCA has solution to make another type of massive updates, say, add a set of lines (without deletion)? We made a script for such cases, though it doesn't have arguments and has to be updated for specific case.
https://itpp.dev/ops/git/github.html

legalsylvain:
@pedrobaeza : thanks for the precision.

tools/massive_change.py
```diff
@@ -0,0 +1,221 @@
...
+
+from __future__ import print_function
+import argparse
+import re
+from . import github_login
+from . import oca_projects
+from .config import read_config
+from github3.exceptions import NotFoundError
+
+DEFAULT_MESSAGE = "[UPD] Massive change [skip ci]"
```
Yajo: Why skip ci? It'd be helpful to know if this commit is guilty for breaking some CI, right?

tools/massive_change.py
```diff
@@ -0,0 +1,221 @@
...
+                        List of specific projects to migrate
+  -e EMAIL, --email EMAIL
+                        Provides an email address used to commit on GitHub if
+                        the one associated to the GitHub account is not public
+  -t TARGET_ORG, --target-org TARGET_ORG
+                        By default, the GitHub organization used is OCA. This
+                        arg lets you provide an alternative organization
+  -m COMMIT_MESSAGE, --message COMMIT_MESSAGE
+                        By default, '[UPD] Massive change [skip ci]', but can
+                        be provided with an alternative one
```
Yajo: IMHO it's better to instruct the user to run `oca-massive-change --help`. Otherwise this will get out of sync pretty soon...

tools/massive_change.py
```diff
@@ -0,0 +1,221 @@
...
+   :alt: Odoo Community Association
+   :target: https://odoo-community.org
+
+This module is maintained by the OCA.
+
+OCA, or the Odoo Community Association, is a nonprofit organization whose
+mission is to support the collaborative development of Odoo features and
+promote its widespread use.
+
+To contribute to this module, please visit https://odoo-community.org.
```
Yajo: Is all this really needed for a script?

tools/massive_change.py
pedrobaeza: By default it should skip it before the change probably won't affect. If you don't like it, just use another, but I prefer this sane default.

tools/massive_change.py
pedrobaeza: This is directly obtained from that output, and I prefer to keep it synced for direct reading.

tools/massive_change.py
pedrobaeza: I prefer to put it.

pedrobaeza:
@yelizariev I think you can already do this with this tool, as you can include newlines with `\n`. Look for example the need that motivates me to develop the option:

```
oca-massive-change 13.0 .travis.yml "pre-commit run --all --show-diff-on-failure\n" "pre-commit run --all --show-diff-on-failure --verbose --color always\n"
```

I'm searching for that part but at the end of the line. Nothing prevents you to put as destination string several lines with that character at the end, and looking for the preceding line including it as well.

tools/massive_change.py
```diff
@@ -0,0 +1,221 @@
...
+            repo.ref('heads/{}'.format(branch.name)).update(commit.sha)
+        return commit
+
+    def _do_change(self, project, path, src, dest, message=None):
+        print("Performing change in project %s/%s" % (self.gh_org, project))
+        repo = self.github.repository(self.gh_org, project)
+        try:
+            repo.branch(self.gh_target_branch)
+        except NotFoundError:
+            print("Branch doesn't exist. Skipping...")
```
joao-p-marques, line 161: Could the target branch be automatically created if it doesn't exist? Something like:
```suggestion
            repo.create_branch_ref(self.gh_target_branch)
```
(See https://github3py.readthedocs.io/en/master/api-reference/repos.html#github3.repos.repo.Repository.create_branch_ref)

tools/massive_change.py
```diff
@@ -0,0 +1,221 @@
...
+            print("Path not found. Skipping...")
+            return
+        new_content = self._replace_content(repo, path, [(src, dest)])
+        if not new_content:
+            print("Source string not found. Skipping...")
+            return
+        tree_data = [new_content]
+        if not message:
+            message = DEFAULT_MESSAGE
+        self._create_commit(repo, tree_data, message)
```
joao-p-marques, line 176: Could the option to, instead of committing directly, create a PR also be added here? https://github3py.readthedocs.io/en/master/api-reference/repos.html#github3.repos.repo.Repository.pull_request

joao-p-marques (commented):
Just some small comments. 

Otherwise, seems very good, specially for solving something like https://github.com/OCA/oca-addons-repo-template/pull/33

tools/massive_change.py
pedrobaeza: This is a changer, not a creator, and it mustn't create anything if not exists. What to populate? There are other tools for such things anyway.

tools/massive_change.py
pedrobaeza: Not, this is not intended for creating PRs. There's no sense on creating 200 PRs, one per repo, for an administrative commit that fixes/updates anything broken. It has to be direct.

tools/massive_change.py
joao-p-marques: I see... Thank you. I was thinking of a wider use case, especially thinking about the PR creation part, but I see it makes no sense.

tools/massive_change.py
```diff
@@ -0,0 +1,221 @@
...
+This script helps to make a massive change across a whole organization
+on contents of a file.
+
+Installation
+============
+
+For using this utility, you need to install these dependencies:
+
+* github3.py library for handling Github calls. To install it, use:
+  `sudo pip install github3.py`.
```
Yajo, line 14: Wrong syntax.

```suggestion
  ``sudo pip install github3.py``.
```

pedrobaeza: Changed

PR #443, opened 2020-02-13
[FIX] gen_addon_readme.template: Avoid trailing whitespace

MiquelRForgeFlow:
### Fixes
- Avoid the fucking trailing whitespace:

Selection_111

### DIFF
--- a/tests/data/readme_tests/addon1/README.expected-acme
+++ b/tests/data/readme_tests/addon1/README.expected-acme
@@ -14,7 +14,7 @@ addon 1
     :target: https://github.com/acme/server-tools/tree/12.0/addon1
     :alt: acme/server-tools
 
-|badge1| |badge2| 
+|badge1| |badge2|
 
 Some description
 
--- a/tests/data/readme_tests/addon1/README.expected-oca
+++ b/tests/data/readme_tests/addon1/README.expected-oca
@@ -20,7 +20,7 @@ addon 1
     :target: https://runbot.odoo-community.org/runbot/149/12.0
     :alt: Try me on Runbot
 
-|badge1| |badge2| |badge3| |badge4| 
+|badge1| |badge2| |badge3| |badge4|
 
 Some description
 
--- a/tests/data/readme_tests/addon_one_maintainer/README.expected-acme
+++ b/tests/data/readme_tests/addon_one_maintainer/README.expected-acme
@@ -14,7 +14,7 @@ addon one maintainer
     :target: https://github.com/acme/server-tools/tree/12.0/addon_one_maintainer
     :alt: acme/server-tools
 
-|badge1| |badge2| 
+|badge1| |badge2|
 
 Some description
 
--- a/tests/data/readme_tests/addon_one_maintainer/README.expected-oca
+++ b/tests/data/readme_tests/addon_one_maintainer/README.expected-oca
@@ -20,7 +20,7 @@ addon one maintainer
     :target: https://runbot.odoo-community.org/runbot/149/12.0
     :alt: Try me on Runbot
 
-|badge1| |badge2| |badge3| |badge4| 
+|badge1| |badge2| |badge3| |badge4|
 
 Some description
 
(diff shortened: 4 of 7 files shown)


MiquelRForgeFlow:
@sbidoul please, could you give me some tips of what should I change to make this green?

sbidoul:
@MiquelRForgeFlow I guess the whitespace between badges is important.

In any case I would wait for another more significant change in the readme generator. Removing a trailing whitespace is not worth regenerating thousands of README files and pushing thousands of packages to PyPI.

Does this trailing space cause issues in practice?

MiquelRForgeFlow:
No, it doesn't cause issues. Just that each time I do a module migration, I see that trailing whitespace in the readmes be deleted and later put again and I thought we could put aside that silliness.

sbidoul:
> I see that trailing whitespace in the readmes be deleted

Do you know what deletes it?

pedrobaeza:
@sbidoul IDEs perform the deletion with the option "remove trailing spaces" activated that lots of people have activated (me included).

MiquelRForgeFlow:
@sbidoul now this PR is green and ready to merge 🚀

sbidoul:
> IDEs perform the deletion with the option "remove trailing spaces" activated that lots of people have activated (me included).

Yes, but don't these IDE do that on save, which you normally don't do on README.rst?

pedrobaeza:
> Yes, but don't these IDE do that on save, which you normally don't do on README.rst?

Yeah, indeed, but sometimes you change a letter in the fragment and don't want to re-run the tool, or you save by accident/habit with no change.

MiquelRForgeFlow:
> you change a letter in the fragment and don't want to re-run the tool

Exactly, I normally run _replace in path_ ("12.0"->"13.0" and "12-0"->"13-0") to avoid re-running the tool.

sbidoul:
I think we'll put gen-addon-readme in the pre-commit config for 13, as Jairo suggested.

MiquelRForgeFlow:
But I don't run pre-commit and don't want to be forced to run that. I prefer running black, isort and such manually.

pedrobaeza:
@MiquelRForgeFlow that's a bit obstinacy: having it integrated in pre-commit is OK and it's more comfortable than doing it manually.

Yes, let's go for that, @sbidoul

MiquelRForgeFlow:
Well, maybe I didn't express correctly. My point is, instead of forcing the user to do more things (like running pre-commit or gen-addon-readme) before pushing, let's automatize more things, creating bots that after merging or pushing create black/isort commits. Instead of a pre-commit thing, let's call a post-merge thing. Let's welcome bots technology instead of manually do things. I don't like running tools, I prefer bots running tools.

pedrobaeza:
pre-commit is for that: for not having to do anything except the initial setup: then everything is done by a bot while you commit, but the "bot" is in your local.

MiquelRForgeFlow:
The bots should be "online", not "local".

pedrobaeza:
I prefer the contrary for reducing the noise of extra commits.

MiquelRForgeFlow:
Extra commits are inevitable. Readme commit, icon commit, translation commit, .pot commit, etc. One more commit for black/isort things doesn't add really much noise.

pedrobaeza:
Generating the README, the icon, etc in your local pre-commit avoids that extra commits. I'm not talking about the "isort, black" one that is performed before the migration for accommodating in the migration the old code. If you see, when you write new code for your migration, black automatically formats it and you include it on your migration commit, not on a new one.

sbidoul:
@MiquelRForgeFlow black/isort is different, because it touches the code. If you have a bot doing extra code formatting commits, git blame becomes a nightmare. So that's clear to me, black stays in pre-commit *and* as a lint check. The rest (readme generator, .pot generator) is optional locally and will continue to be done by the bots.

MiquelRForgeFlow:
Well, ok, I was just expressing my point of view. But let's leave it here, all this talk is outside of the topic of this PR.

dreispt:
Can the READMEs be regenerated only for the modules involved in the PR?
We are already doing that for linting, right?

sbidoul:
Since ocabot merge seems to be widely used, we could regenerate addons READMEs only when doing ocabot merge, and remove it from the nightly jobs. Although there are still people merging manually here and there.

MiquelRForgeFlow:
> we could regenerate addons READMEs only when doing ocabot merge, and remove it from the nightly jobs

It seems to me we could do this nowadays, right?

PR #444, opened 2020-02-16
[IMP] Add stale tags

alexey-pelykh:
@pedrobaeza going to use these two tags in a PR for `oca-github-bot` to detect, warn, and auto-close stale issues and PRs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@ sudo: false
 language: python
 
 python:
-  - "2.7"
   - "3.5"
   - "3.6"
 
--- a/tools/set_repo_labels.py
+++ b/tools/set_repo_labels.py
@@ -25,6 +25,7 @@ all_labels = {
     'wontfix': 'ffffff',
     'migration': 'd4c5f9',
     'ready to merge': 'bfdadc',
+    'stale': '555555',
 }
 
 


.travis.yml
```diff
@@ -2,7 +2,6 @@ sudo: false
 language: python
 
 python:
-  - "2.7"
```
sbidoul: I'm not opposed to dropping python 2.7, but this should be in another PR, that also adds [python_requires](https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires) in `setup.py`.

yvaucher:
ping @alexey-pelykh can you separate both changes as asked by @sbidoul ?

alexey-pelykh:
I'm just not sure if this PR is relevant, as @sbidoul was testing GitHub bots to detect stale PRs and things.

sbidoul:
In any case having a stale label can't hurt :)

I've been experimenting with that indeed: https://github.com/OCA/mis-builder/blob/10.0/.github/workflows/stale.yml

It did work fine in my simple test. It would need to be replicated to all repos with a tool that is still to be developed (https://github.com/OCA/oca-github-bot/pull/85#issuecomment-605819787).

The worry I have with GitHub such action is how closer we'd be to GitHub API rate limits. 5000 requests/hours is not that much for 200 repos.

pedrobaeza:
What about existing bots?

sbidoul:
> What about existing bots?

That too, yes. We have a balance to find between what we do with our own bot (lots of specialized things can only be down by our own), and reuse existing actions, at the risk of complexifying the architecture.

pedrobaeza:
The rate limit is worrying, so I would say that this stale feature should be handled by GitHub bots that don't consume this rate.

sbidoul:
This was finished in https://github.com/OCA/maintainer-tools/pull/503

PR #447, opened 2020-04-19
Use the python3 way to create virtualenv

eLBati:


README.md
```diff
@@ -7,7 +7,7 @@
 
     $ git clone git@github.com:OCA/maintainers-tools.git
     $ cd maintainers-tools
-    $ virtualenv env
+    $ python3 -m venv env
```
sbidoul, line 10: Note that to get something really equivalent you'll need to add `env/bin/pip install -U pip setuptools wheel` to get the latest `pip` and `setuptools` and more importantly to add `wheel`. Indeed python's venv module does not install `wheel` and when it is absent, pip downgrades to legacy behaviour in several code paths.

dreispt: This means that `wheels` installation is also needed?

sbidoul: @dreispt it's not strictly necessary but when it's absent, pip falls back to legacy behaviours. So better take good habits.

yvaucher:
@eLBati note that the README contains 2 places where the git clone and env are done:

You will want to sync the procedure here:
https://github.com/OCA/maintainer-tools/blob/master/README.md#developers

sbidoul:
In any case I'd say it's still better to recommend virtualenv which gives better defaults for python packaging this days. Moreover python3 venv is not even installed by default on Debian and Ubuntu due to the unbundling of the python stdlib done by the Debian distro.

README.md
```diff
@@ -7,7 +7,7 @@
...
     $ . env/bin/activate
     $ python setup.py install
```
simahawk, line 12: BTW I'd use `pip install -e .`

simahawk (commented):
What do we do w/ this?

README.md
sbidoul: Now that pip does not do `setup.py install` anymore, I'm now comfortable with recommending `python3 -m venv` followed by a pip upgrade.

dreispt:
I think this can be closed.

PR #450, opened 2020-04-30
[FIX] gen-addon-readme: generate badges with secure links

skeller1:
### Milestone (Odoo version)
- 

### Module(s)
- gen-addon-readme: generate readme with secure badge links

### Fixes / new features
- make license links secure (using https) in all repository files (like we do it for addon migrations)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,4 +1,4 @@
-# License AGPLv3 (http://www.gnu.org/licenses/agpl-3.0-standalone.html)
+# License AGPLv3 (https://www.gnu.org/licenses/agpl-3.0-standalone.html)
 import os
 import setuptools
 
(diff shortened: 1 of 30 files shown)


skeller1:
> Thanks for the contrib. The update to the headers looks fine.
> 
> However, changing the readme generator will cause a regeneration of all readmes on all repos, which will in turn re-publish everything to PyPI. This is not worth the noise, IMO, especially since `www.gnu.org` has an HSTS policy which redirects to https anyway:
> 
> ```
> $ wget -O - http://www.gnu.org/licenses/agpl-3.0-standalone.html > /dev/null
> URL transformed to HTTPS due to an HSTS policy
> --2020-04-30 18:04:27--  https://www.gnu.org/licenses/agpl-3.0-standalone.html
> Resolving www.gnu.org (www.gnu.org)... 2001:470:142:3::a, 209.51.188.148
> Connecting to www.gnu.org (www.gnu.org)|2001:470:142:3::a|:443... connected.
> HTTP request sent, awaiting response... 200 OK
> ```

Okay, removed the changes to https from `gen-addon-readme` script.

Thank you

sbidoul:
Thanks!

PR #455, opened 2020-05-29
[FIX] README template section levels

dreispt:
The README generation could return an error like this:
```
  docutils.utils.SystemMessage: account_avatax/README.rst:349: (SEVERE/4)
  Title level inconsistent:

  Authors
  ~~~~~~~
```
In fact the Authors section is inside a Credits first level title, ``====``,
and should be a second level title, ``----``,
instead of a third level, ``~~~``.


sbidoul:
Do you have an example where it failed?

dreispt:
Yes: 
Pull https://github.com/OCA/account-fiscal-rule/pull/116
Then run ``oca-gen-addon-readme --repo-name=account-fiscal-rule --branch=13.0 --addon-dir=account_avatax``

sbidoul:
@dreispt I think it should work if you change the section underlines to tilde (`~`) in your fragments.
It has been a restriction since the beginning that this is the only allowed title underline in fragments.
It is mentioned in the example [here](https://github.com/OCA/maintainer-tools/blame/c57170038fb502b86e694b5137b9e051c1c22754/template/module/readme/HISTORY.rst#L6-L9), although admitedly it is not easily discoverable.

dreispt:
Indeed those instructions could/should be in a better place. DESCRIPTION.rst sounds like the best place for it. Not finding them, I referred to the RST docs, and that led to this...

PR #457, opened 2020-06-03
[ADD] pre-commit hook to exclude not installable addons

sbidoul:
In repos with `installable=False` addons it is useful to exclude these from the pre-commit checks.

This PR provides a pre-commit hook that updates `.pre-commit-config.yaml` with the exclude list, making sure the exclude list remains up-to-date.

Typical usage in https://github.com/OCA/oca-addons-repo-template/pull/3


sbidoul:
@ThomasBinsfeld

lmignon (approved):

PR #460, opened 2020-06-10
[IMP] readme template: subsection instructions

dreispt:
Explain how to add sections and subsections in README fragments.

### DIFF
--- a/template/module/readme/DESCRIPTION.rst
+++ b/template/module/readme/DESCRIPTION.rst
@@ -3,3 +3,6 @@
 This module extends the functionality of ... to support ...
 and to allow you to ...
 
+Fragments may contain section titles. The first title level must be underlined
+with the ``~`` (tilde) character. Subsections can be underlined using any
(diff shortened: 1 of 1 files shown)


template/module/readme/DESCRIPTION.rst
```diff
@@ -3,3 +3,6 @@
 This module extends the functionality of ... to support ...
 and to allow you to ...
 
+Fragments may contain section titles. The first title level must be underlined
+with the ``~`` (tilde) character. Subsections can be underlined using any
+non-alphanumeric character, other than ``~``, ``=`` and ``-``.
```
pedrobaeza, line 8: Examples of these allowed non-alphanumeric characters?

sbidoul:
@dreispt thank you for clarifying this.

I think if we want to allow more than one level in fragments we need to say explicitly which underline character to use. Otherwise we risk blocking the evolution of other auto generated sections, or prevent the combination of several readmes in a single documents.

dreispt:
@dreispt OK, I see the point, I will propose levels.

PR #464, opened 2020-06-18
Improvements of tests

rousseldenis:
In order to avoid too much travis jobs (and getting slow on other repos), I propose to set three steps for tests. The advantage is that the folowing ones are not running if preceding ones are failing.

One for 'lint', one for odoo based tests and one for OCB based ones (if odoo tests are failing, there is no reason of running OCB ones)

@sbidoul @pedrobaeza

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,23 @@ install:
     - pip install tox-travis
     - pip install coveralls
 
+# stages:
+#  - linting
+#  - test
+#  - test-ocb
+
+#jobs:
+#  include:
+#    - stage: linting
+#      env:
+#        - LINT_CHECK="1"
+#    - stage: test
+#      env:
+#        - TESTS="1" ODOO_REPO="odoo/odoo"
+#    - stage: test-ocb
+#      env:
+#        - TESTS="1" ODOO_REPO="OCA/OCB"
+
 script:
     - tox
 


pedrobaeza:
This doubles more or less the end of the tests, and OCB tests must wait to Odoo ones if I'm understanding correctly.

rousseldenis:
> This doubles more or less the end of the tests, and OCB tests must wait to Odoo ones if I'm understanding correctly.

Yes, but that allows more time for other ones and dispatch more equally the jobs through all repos (eg: repos with plenty of exclude/include loads a lot of jobs at once and block other ones)

rousseldenis:
But at least, I would do two stages : lint and test like in https://github.com/OCA/server-tools/pull/1850

sbidoul:
Yes, it's hard to tell if we'll win with this. I'd say it can help, because most of the time jobs don't start because we are at maximum capacity, and there are many double failures (odoo+ocb), so this may reduce the total number of jobs to run.

pedrobaeza:
Yeah, for failing jobs, indeed we gain, but for example for ocabot merge jobs, that are likely to be green, we get more time till the merge.

simahawk:
We are using GH actions now. Sounds like we can close this, right?

PR #465, opened 2020-06-23
[IMP] Allow to set a custom icon

ivantodorovich:
# oca-gen-addon-icon

- Allows to use a custom icon with `--src-icon` option.

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,6 +127,10 @@ To generate the icon for the module `auth_keycloak`:
 
     $ oca-gen-addon-icon --addon-dir=auth_keycloak
 
+A custom icon can be added using the `--src-icon` argument:
+
+    $ oca-gen-addon-icon --addon-dir=auth_keycloak --src-icon=/path/to/custom/icon.png
+
 
 ### Auto fix pep8 guidelines
 
--- a/tools/gen_addon_icon.py
+++ b/tools/gen_addon_icon.py
@@ -17,16 +17,20 @@ ICON_TYPE = 'png'
 ICON_TYPES = ['png', 'svg']
 
 
-def gen_one_addon_icon(icon_dir, filetype=ICON_TYPE):
+def gen_one_addon_icon(icon_dir, src_icon=None, filetype=ICON_TYPE):
     icon_filename = os.path.join(icon_dir, 'icon.%s' % filetype)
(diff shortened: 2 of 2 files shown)


GSLabIt:
@ivantodorovich src_path could be an url?

ivantodorovich:
> @ivantodorovich src_path could be an url?

Hi @GSLabIt !

It has to be a local path. It could definitely be improved to accept urls, but some extra code would be required to identify them and to download the remote image.

However this PR didn't get much attention :) I'm using this in our processes though

pedrobaeza:
See #491 as well

ivantodorovich:
> See #491 as well

That is indeed a very cool PR, but with a different purpose.

This one is to be used by non-OCA CI tools to add a default module icon to your company modules: 
image

By allowing to set a custom icon, you can use this very same tool in your company repositories, using another icon like:
image

It's of no use for the OCA, but it's usefull for people using this tool in other repositories

pedrobaeza (approved):

GSLabIt:
> > See #491 as well
> 
> That is indeed a very cool PR, but with a different purpose.
> 
> This one is to be used by non-OCA CI tools to add a default module icon to your company modules:
> image
> 
> By allowing to set a custom icon, you can use this very same tool in your company repositories, using another icon like:
> image
> 
> It's of no use for the OCA, but it's usefull for people using this tool in other repositories

any hint on how to use in oct-bot? @pedrobaeza

pedrobaeza:
To use what?

GSLabIt:
> To use what?

`--src-icon` to use a different icon on oca-bot? i think it can be passed as extra param, but not sure path to pass in order to make it work properly 😬

GSLabIt (approved):

pedrobaeza:
On OCA bot, there's no sense to use another icon except OCA one.

GSLabIt:
> On OCA bot, there's no sense to use another icon except OCA one.

i agree, just asking because we deployed oca-bot our company. Thx anyway.

pedrobaeza:
Then propose a PR in such project for customizing that part.

GSLabIt:
> Then propose a PR in such project for customizing that part.

I'll try, thx

PR #466, opened 2020-06-25
Remove _view suffix from views

Yajo:
Views are already in `views` folder, so I think this is redundant.


pedrobaeza:
I don't, as when searching for files, you don't see what is (data, demo or view). Most IDEs also allow to search by file name, and the path is secondary, so I prefer the Python "explicit is better than implicit".

:-1:

Yajo:
That would apply for any kind of file, not just views. Then, we'd have to name demo data `demo/res_partner_demo.xml`, then other data `data/res_partner_data.xml` and templates `templates/res_partner_templates.xml`.

In this gif you can see how a modern IDE such as VSCode doesn't need this `_views` suffix at all (you can see it highlights `view` from the path and `res_partner` from the file name). You can also see how many people are already ignoring that guideline organically; even there's one doing it the opposite: `view_res_partner.xml`.

Peek 04-08-2020 08-46

Also, removing the repeated word in the file path adheres to the DRY principle.

So still :+1: here.

>  Please rename also the `wizard_model_view.xml` file in https://github.com/OCA/maintainer-tools/tree/master/template/module/wizards

Fixed.

pedrobaeza:
As said, I don't like this, so :-1: for me. Your example is very specific, as you don't have long paths for it. I don't see the final folder name as my path is longer than the view.

Yajo:
Well, OK, that's +2-1=+1. Let's wait until other members vote. Thanks @pedrobaeza for your opinion!

bealdav (approved):

pedrobaeza (changes requested):
Explained in previous comments

pedrobaeza:
Another thing to take into account: inside `views` folder, you can have QWeb templates for reports for the model, menus, etc, so file naming is important to highlight this: `account_move_views.xml`, `report_account_move.xml`, `account_move_menu.xml`, etc.

PR #467, opened 2020-07-03
Add pipx method to installation methods

gdgellatly:
Current installation method refers to using virtualenv, but with pipx gaining in popularity we should add pipx install instructions.


sbidoul:
Thanks!

PR #471, opened 2020-10-04
Add oca-create-migration-issue

sbidoul:
This is what I used to create the migration issues for 14.0.

Extracted from @pedrobaeza 's `oca-migrate-branch` and `oca-migrate-branch-empty`.


pedrobaeza:
Isn't better to add a flag to existing scripts? Alternatively, we can remove this piece of code from the other place and let only one place for it.

sbidoul:
> Alternatively, we can remove this piece of code from the other place and let only one place for it.

done

pedrobaeza (approved):

PR #486, opened 2020-12-12
[ADD] oca-copier-update

sbidoul:
This is the script I currently use to update dotfiles. I post it here in draft state case anyone is interested.

Needs thinking:
- This should probably be a bot action
- It runs pre-commit locally. If ok, it pushes to the main branch with [ci skip]. If not it creates a PR ([example](https://github.com/OCA/event/pull/200)). This is a shortcut to save CI build time, but in the general case, a dotfiles update could break something else than pre-commit. So it might be best to always create a PR and immediately ocabot merge it. The drawback is that it will create a ci build storm...
- Also, running pre-commit locally might have security implications
- Yet, it is currently necessary to autoformat copier-generated files (e.g. yaml files)

### DIFF
--- a/tools/copier_update.py
+++ b/tools/copier_update.py
@@ -99,8 +99,8 @@ def _iterate_repos_and_branches(repos: str, branches: str) -> Iterable[Tuple[str
 @click.option("--org", default="OCA")
 @click.option("--repos", required=True)
 @click.option("--branches", required=True)
-@click.option("--git-user-name")
-@click.option("--git-user-email")
+@click.option("--git-user-name", default="oca-git-bot")
+@click.option("--git-user-email", default="oca-git-bot@odoo-community.org")
 @click.option("--skip-ci/--no-skip-ci", default=False)
 def main(
     org: str,
--- a/tools/copier_update.py
+++ b/tools/copier_update.py
@@ -1,7 +1,6 @@
 """Run copier update on a branch in all addons repos.
 """
 from pathlib import Path
-import os
 import subprocess
 import textwrap
 from typing import Optional
@@ -25,13 +24,6 @@ def _make_commit_msg(ci_skip: bool) -> str:
     if ci_skip:
         msg += " [ci skip]"
     return msg
-
-
-def _has_rej_file() -> bool:
-    for _, _, filenames in os.walk("."):
-        if any(f.endswith(".rej") for f in filenames):
-            return True
-    return False
 
 
 def _get_update_dotfiles_open_pr(org: str, repo: str, branch: str) -> Optional[str]:
@@ -92,6 +84,8 @@ def _make_update_dotfiles_pr(org: str, repo: str, branch: str) -> None:
 @click.argument("branch")
 def main(branch):
     for repo in get_repositories():
+        if repo not in ("event",):
+            continue
         try:
             with temporary_clone(repo, branch):
                 print("=" * 10, repo, "=" * 10)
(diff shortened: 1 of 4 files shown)


tools/copier_update.py
```diff
@@ -0,0 +1,123 @@
...
+        try:
+            with temporary_clone(repo, branch):
+                print("=" * 10, repo, "=" * 10)
+                # set git user/email
+                subprocess.check_call(
+                    ["git", "config", "user.name", "oca-git-bot"],
+                )
+                subprocess.check_call(
+                    ["git", "config", "user.email", "oca-git-bot@odoo-community.org"],
+                )
```
Yajo, line 96: This might not be the best if you're running locally. Maybe you can just add `--author='oca-git-bot <oca-git-bot@odoo-community.org>'` to `git commit`?

tools/copier_update.py
```diff
@@ -0,0 +1,123 @@
...
+                subprocess.check_call(["git", "add", "."])
+                # run up to 3 pre-commit passes, in case autofixers
+                # (which cause pre-commit to fail when they change files)
+                # resolve issues
+                for _ in range(3):
+                    r = subprocess.call(["pre-commit", "run", "-a"])
+                    # git add, in case pre-commit created new files
+                    subprocess.check_call(["git", "add", "."])
+                    if r == 0:
+                        break
```
Yajo, line 115: I think this part should be a little bit improved.

There's 2 things here that are merged but should be IMHO split.

The 1st thing is to reformat dotfiles themselves. This is desired as they will almost 100% of the time will fail (beause `.copier-answers.yml` has to be reformated always).

The 2nd thing is to reformat all the rest of source code.

If you're migrating a module and extract the git history from older branch, and then find in between commits labeled like `[IMP] update dotfiles`, but the real change is not in dotfiles but in source code, that can be a little misleading. Besides, having a clear separation between what's updated by copier and what by pre-commit (in other parts) is also helpful to later debug if there's a problem in the template or in pre-commit configuration.

So, I think it would be better like this (pseudocode):

```py
failed = False

# Update from copier and reformat its changes
_run_copier()
for n in range(3):
    try:
        pre_commit("install")
        git("add", ".")
        git("commit", "-m", "[IMP] update dotfiles")
    except ProcessError:
        if n == 2:
            failed = True
            git("commit", "-m", "[IMP] update dotfiles", "--no-verify")

# Reformat all other code if needed
try:
    pre_commit("run", "-a")
except ProcessError:
    git("add", ".")
    try:
        git("commit", "-m", "[IMP] pre-commit run")
    except ProcessError:
        git("commit", "-m", "[IMP] pre-commit run", "--no-verify")
        failed = True

# Open PR because it failed
if failed:
    _make_update_dotfiles_pr(ORG, repo, branch)
```

sbidoul: The more I think to it the more I'm convinced that prettier should not touch files generated by the copier template.
For instance this can cause copier merge conflicts (.rej) as I have had while doing https://github.com/OCA/server-tools/pull/2203, due to the line break added by prettier behind copier's back.

pedrobaeza: I agree

joao-p-marques: It's true, and it might be a valid approach. However, with copier and jinja templating it will always be hard to generate "pretty" and easy to read configuration files, so we can always benefit from pre-commit formatting.

OTOH, it might be that the user edits the file manually, and in a copier update an actual conflict is created, and a `.rej` file is generated. In that case, having pre-commit checking all the files in the repo might be useful... 

Personally, I have gotten pretty used to the flow of (1) `copier update` (2) Run pre-commit on those generated files (3) Fix some conflicts if needed and (4) commit the changes. However, maybe @Yajo can give some more insight on the pros and cons of each option.

Yajo: It's very unlikely that the pre-commit part is guilty of some conflict, as that's supported upstream by copier. I mean that, when doing the previous copy to compare with current state, copier already attempts a 2nd commit if the 1st fails, assuming that in [tasks](https://copier.readthedocs.io/en/stable/configuring/#tasks) you could be doing a `pre-commit install` or equivalent.

That said, another point to take into account is that any copier update can usually mean an update to the pre-commit configuration, which itself could lead to violations of linters/formatters, even in other files. Of course I'm talking of other projects not in OCA, where we freeze that part as much as possible per version.

At the end of the day, it's just a criteria. You can opt for not formatting those files, saving yourself from an extra step, or you can opt for formatting those, producing beautier configs always. I opted for the 2nd, but if you all agree we can use the 1st, it's not a problem for me.

Yajo:
These days I've been thinking about another approach.

There could be a copier GH action. It could run weekly and attempt to update using copier. This way, once you do a new template release, the bot will take care of upgrading. It could open a PR or do a direct push if all succeeds.

The action could be in https://github.com/copier-org directly, as this isn't something specific for OCA.

joao-p-marques:
> These days I've been thinking about another approach.
> 
> There could be a copier GH action. It could run weekly and attempt to update using copier. This way, once you do a new template release, the bot will take care of upgrading. It could open a PR or do a direct push if all succeeds.
> 
> The action could be in https://github.com/copier-org directly (I have permissions to create there), as this isn't something specific for OCA.

I think this could be really cool :+1: It could even run more frequently and start by checking for new releases in the template repository before even attempting the update to reduce overhead on the scheduled action.

sbidoul:
I'm merging this since this is what I use to do the mass updates.

PR #495, opened 2021-04-14
[IMP] Version-specific website tour code samples

Yajo:
This specific part of odoo changes a bit depending on your Odoo version.

Reflecting it in the template will help when developing a new addon.

cc @chienandalu

### DIFF
--- a/template/module/tests/test_something.py
+++ b/template/module/tests/test_something.py
@@ -28,10 +28,6 @@ class SomethingCase(TransactionCase):
 
 
 class UICase(HttpCase):
-
-    post_install = True
-    at_install = False
-
     def test_ui_web(self):
         """Test backend tests."""
         self.phantom_js(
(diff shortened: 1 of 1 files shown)


chienandalu (approved):
:+1:

pedrobaeza:
CI is failing. I wouldn't change the method name, but put the proper one, and prefix a comment explaining which one for each version.

Yajo:
So, your suggestion is that I create 3 methods with the same name?

pedrobaeza:
Yes. If not, people may think that the method name is that one.

Yajo:
Fixed.

pedrobaeza:
I still think creating 3 methods with the same name is the correct thing, as:

- It's easier to delete non proper ones.
- Linters will complain if you leave several of them.
- Comment about the usage can be put as docstring and more easily identifiable.

Yajo:
Done

pedrobaeza (approved):
I don't know though why CI is red.

Yajo:
It seems some misconfiguration with coveralls. But this has nothing to do with CI, right? This is just dead code to copy and paste elsewhere, I don't think those ❌ are important here.

pedrobaeza:
Well, we should fix it. I think @sbidoul fix a similar thing recently. Do you have a clue?

Yajo:
Yes, although that is obviously a totally different issue. I think we can merge this.

pedrobaeza:
Well, although unrelated indeed, merging something red is not usual, so I let @sbidoul to decide about that.

sbidoul:
I fixed the coveralls issue in #497

PR #501, opened 2021-06-14
copy_maintainer: fix invite

sbidoul:
The invite method does not exist in github3.py version 2.0


sbidoul:
This is deployed @gurneyalex

PR #503, opened 2021-07-28
[IMP] oca-set-repo-labels: add new labels and allow to set descriptions

ivantodorovich:
Following up on https://github.com/OCA/oca-addons-repo-template/pull/63#issuecomment-887637020, and also adding label descriptions

_NOTE: Sort of did it blindly without testing as I'm not maintainer to the repos_

### DIFF
--- a/tools/set_repo_labels.py
+++ b/tools/set_repo_labels.py
@@ -10,51 +10,69 @@ from .github_login import login
 REPO_TO_IGNORE = [
 ]
 
-# here is the list of labels we need in each repo
-all_labels = {
-    'bug': 'fc2929',
-    'duplicate': 'cccccc',
-    'enhancement': '84b6eb',
-    'help wanted': '159818',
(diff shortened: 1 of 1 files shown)


Yajo (approved):
Also I approve blindly.

sbidoul:
@ivantodorovich I had to make https://github.com/OCA/maintainer-tools/commit/ac8d3aff67522471ed7ac3cb998b883850ad2ae3 to fix it (partially).

But the `label.update` is return a 404...

ivantodorovich:
Thanks @sbidoul !

I tested locally, changing a bit the code so that it works on my own repositories.
It's working but only with `github3.py >= 1.2`, so we should bump this library

I did not get 404 error in any of my tests, even with previous github3 versions, so I'm not sure what the issue could be.. I did confirm that it's not working with `github3.py == 1.0`, though, as it doesn't support setting label descriptions

Issue #454, opened 2020-05-22
[RFC][IMP] gen_addons_table : Add maintainers and maturity columns


legalsylvain:
Hi all, 

I was thinking about the possiblity to update the gen_addons_table script, to add two new columns : 

- a Maintainers column with the list of the maintainers as set in each module readme file. ([Sample](https://github.com/OCA/mis-builder/tree/10.0/mis_builder#maintainers))
- a badge column, with the maturity level (alpha, beta, ...) as set also in each module readme file([Sample](https://github.com/OCA/mis-builder/tree/10.0/mis_builder#mis-builder))

What do you think ? 

It could help to : 
- have an overview of the unmaintained modules ; 
- have an overview of the modules maturity ;

thanks for your feedback. CC : @sbidoul, @pedrobaeza

pedrobaeza:
As you know, I don't like too much that table, as it adds a lot of noise in commits and require the scanner that updates it, but if you have it, the maintainer can be interesting (if updated, which not always is true), and the maturity is as well, but people doesn't care too much about this unfortunately, so most of them will show beta.

sbidoul:
Hi Sylvain, I think it's a good idea to give a better visibility to these two important concepts. I'd put the maturity badge inside the version cell to save some space.

PR #505, opened 2021-08-01
[IMP][RFC] oca-gen-addons-table : add developpement_status badge.

legalsylvain:
- Add maintainers column in the readme file of the repository.

The inserted image contains the following code 
``[MAINTAINER](https://github.com/MAINTAINER)``

- Add maturity badge in version column. 

**Result for the OCA/pos repository :**

image

Full result : https://github.com/legalsylvain/pos/tree/12.0-TEST-oca-gen-addons-table-maintainers

Thanks for your review.

CC : original author / commiter of this file : @sbidoul, @StefanRijnhart.
CC : @quentinDupont 

Closes: #454

### DIFF
--- a/tests/test_repo/README.md.expected
+++ b/tests/test_repo/README.md.expected
@@ -11,17 +11,17 @@ ${REPO_DESCRIPTION}
 
 Available addons
 ----------------
-addon | version | summary
---- | --- | ---
-[module1](module1/) | 8.0.1.0.0 | Module 1 summary
+addon | version | maintainers | summary
+--- | --- | --- | ---
+[module1](module1/) | 8.0.1.0.0 [![Production/Stable](https://img.shields.io/badge/maturity-Production%2FStable-green.png)](https://odoo-community.org/page/development-status) | [![sbidoul](https://github.com/sbidoul.png?size=30px)](https://github.com/sbidoul) | Module 1 summary
 
 
 Unported addons
 ---------------
-addon | version | summary
---- | --- | ---
-[module3](__unported__/module3/) | 8.0.1.0.0 (unported) | Module 3 summary
-[module2](module2/) | 8.0.1.0.0 (unported) | Module 2 summary
+addon | version | maintainers | summary
+--- | --- | --- | ---
+[module3](__unported__/module3/) | 8.0.1.0.0 (unported) [![Beta](https://img.shields.io/badge/maturity-Beta-yellow.png)](https://odoo-community.org/page/development-status) | | Module 3 summary
+[module2](module2/) | 8.0.1.0.0 (unported) [![Beta](https://img.shields.io/badge/maturity-Beta-yellow.png)](https://odoo-community.org/page/development-status) | | Module 2 summary
 
 [//]: # (end addons)
 
--- a/tests/test_repo/module1/__openerp__.py
+++ b/tests/test_repo/module1/__openerp__.py
@@ -8,6 +8,8 @@
     "category": "Uncategorized",
     "website": "https://odoo-community.org/",
     "author": "<AUTHOR(S)>, Odoo Community Association (OCA)",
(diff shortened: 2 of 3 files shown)


sbidoul:
CI is red, but otherwise :+1:

rousseldenis:
Seems great indeed. +1 too

legalsylvain:
Hi @sbidoul and @rousseldenis.

After thinking again regarding addons table, I'm not very confortable with the result of having a "Beta" value set by default. So : 
I split the two features : 
- the maintainers colum feature is done here : https://github.com/OCA/maintainer-tools/pull/506. I fixed tests, and i think it can be merged quickly
- Regarding status column I don't think that the current default "Beta" value is a good thing because we have no idea regarding the quality of the module. For the time being It is displaying 99% of the OCA module as "Beta" and it's not True. 

So, regarding that second point, i propose to display a bagde in the addons-table AND in the README file,ONLY if a value is explicitily set.
If you're ok, I'll update this PR, to patch also ``gen_addon_readme.py`` to remove badge if there is no value set in the ``developpement_status`` key.

rousseldenis:
> Hi @sbidoul and @rousseldenis.
> 
> After thinking again regarding addons table, I'm not very confortable with the result of having a "Beta" value set by default. So :
> I split the two features :
> 
> * the maintainers colum feature is done here : [[IMP] add maintainers column in addons table. #454 #506](https://github.com/OCA/maintainer-tools/pull/506). I fixed tests, and i think it can be merged quickly
> * Regarding status column I don't think that the current default "Beta" value is a good thing because we have no idea regarding the quality of the module. For the time being It is displaying 99% of the OCA module as "Beta" and it's not True.
> 
> So, regarding that second point, i propose to display a bagde in the addons-table AND in the README file,ONLY if a value is explicitily set.
> If you're ok, I'll update this PR, to patch also `gen_addon_readme.py` to remove badge if there is no value set in the `developpement_status` key.

Yes, that will help finding modules that don't have development_status. +1

sbidoul:
> After thinking again regarding addons table, I'm not very confortable with the result of having a "Beta" value set by default.

That is the reality of the default value, though. It might be useful to make it more visible so people think about it.

legalsylvain:
Hi @sbidoul : 
> That is the reality of the default value, though

I understand your point. my though is just that there is a difference between "Beta" and "undefined". For the time being, 99% of the OCA has no development status defined. and there are not all in the "beta" status.
So, I think that letting blank if it's undefined and display real development status could be a great improvement. 

Alternatively, we can use a "Unknown" Status like this : 

image

sbidoul:
The point is that it is not unknown. All our tooling considers it is Beta. Beta is displayed in the README, the [GitHub actions will test status compatibility](https://github.com/OCA/oca-addons-repo-template/pull/61/files#diff-1a1a06e77eea019f9a493c0ad1120d920d4c389fca8d74e7df059780d2b963afR82), etc.

So making this more visible should encourage people to explicitly define the status they want.

sbidoul:
So you could show the Beta badge *plus* something else that says it came by default.

legalsylvain:
> The point is that it is not unknown. All our tooling considers a missing status is Beta. Beta is displayed in the README, the GitHub actions will test status compatibility using Beta as default, they are published as Beta on PyPI, etc.

> So making this more visible should encourage people to explicitly define the status they really want.

My proposal is not to change github actions, pypi, etc... But just to avoid to display in the readme of the module, and the general readme of the repo, something wrong. (or automatically filled).

> So you could show the Beta badge plus something else that says it came by default.

Indeed. could be the solution. Even if i don't it know if it will be understandable.

Nevermind ! let's put this topic to rest and let other people express themselves on this subject.

could you review consensual #506 in the meantime ?

sbidoul:
@legalsylvain what do we do with this ?

PR #514, opened 2021-10-21
[RFR] OpenUpgrade is an addons repo now

StefanRijnhart:
### Milestone (Odoo version)
- 16.0

### Module(s)
- 

### Fixes / new features
- Allow new series branches on the OpenUpgrade project to be created automatically

### DIFF
--- a/tools/config.py
+++ b/tools/config.py
@@ -57,7 +57,6 @@ NOT_ADDONS = {
     'odoo-sentinel',
     'odoo-sphinx-autodoc',
     'odoorpc',
-    'OpenUpgrade',
     'openupgradelib',
     'pylint-odoo',
 }


pedrobaeza:
But as said on the issue, this can't be initialized the same way as the rest. It has its own specific CI.

sbidoul:
Ok, then better leave it alone.

StefanRijnhart:
Oh ok.

legalsylvain:
> Now that I have prepared 15.0 branch, definitively this can't be templated, as there are very specific changes corresponding to OpenUpgrade, so it's worse to start for the template and change that things, that copying previous branch, and adapt it:

Thanks @pedrobaeza for this work !

PR #515, opened 2021-10-21
[IMP] re-enable _make_default_branch

jjscarafia:
Now it's working properly


sbidoul:
> There are still people using them, so it's worth to re-add it, but indeed the new one should be covered as well.

I'm not sure I agree with that. This project is dedicated to tools for OCA maintenaners, and keeping stuff we don't use for OCA adds unnecessary maintenance burden. People are free to pickup the code in other projects.

simahawk:
Is this still relevant?

pedrobaeza:
I think so, as we still have repos with 13.0 or 14.0 as default branch, and I have to ping you got changing them.

Issue #524, opened 2022-02-21
[gen-addon-readme] Add support of runboat


rousseldenis:
@sbidoul @gurneyalex

sbidoul:
Yes, you can do a PR if you like.

This depends on #520 for deployment.

pedrobaeza:
Well, basically all READMEs will change, so no need of #520, isn't it?

sbidoul:
There is no strong need to regenerate all READMEs since the redirection to runbot is in place.
And we still need #520 to avoid a regeneration of everything each time we want to tweak the README for any reason .

pedrobaeza:
OK, I prefer to have the new shiny badge "Try me on runboat", but as you prefer.

sbidoul:
This will be done as part of #564

PR #533, opened 2022-08-03
[IMP] runbot_with_ids.txt: Fix typo for realestate

rousseldenis:
@simahawk Missed this one for gen-addon-readme

### DIFF
--- a/tools/repos_with_ids.txt
+++ b/tools/repos_with_ids.txt
@@ -132,7 +132,7 @@
 234|github.com/OCA/connector-lengow
 237|github.com/OCA/l10n-vietnam
 238|github.com/OCA/l10n-thailand
-239|github.com/OCA/vertical-realstate
+239|github.com/OCA/vertical-realestate
 240|github.com/OCA/maintenance
 248|github.com/OCA/mis-builder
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Shouldn't we update the README template for using new runboat links?

simahawk:
The fix is legit but... do we still need this file at all? There's no "runbot" anymore for OCA.

rousseldenis:
> The fix is legit but... do we still need this file at all? There's no "runbot" anymore for OCA.

Yes. But this is to restablish a sane situation.

gen-addon-readme does not work anymore for that repository as it relies on this file.

I don't know the real implementation of the tool, maybe something ha to be done there too

rousseldenis:
There is already this : https://github.com/OCA/maintainer-tools/issues/524

rousseldenis:
> Shouldn't we update the README template for using new runboat links?

@pedrobaeza Should be done there : https://github.com/OCA/maintainer-tools/blob/master/tools/gen_addon_readme.py#L202

As template is interpreting 'badges' variable.

rousseldenis:
Done there : #534

simahawk:
Cool, thanks!

PR #534, opened 2022-08-03
[IMP] gen_addon_readme: Don't rely anymore on runbot id/generate runboat link

rousseldenis:
Fixes :

- #524

### DIFF
--- a/tools/gen_addon_readme.py
+++ b/tools/gen_addon_readme.py
@@ -14,7 +14,6 @@ from jinja2 import Template
 
 from .gitutils import commit_if_needed
 from .manifest import read_manifest, find_addons, NoManifestFound
-from .runbot_ids import get_runbot_ids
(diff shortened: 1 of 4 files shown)


tools/gen_addon_readme.py
```diff
@@ -14,7 +14,6 @@
 
 from .gitutils import commit_if_needed
 from .manifest import read_manifest, find_addons, NoManifestFound
-from .runbot_ids import get_runbot_ids
```
simahawk, line 17: can't we trash related files?

rousseldenis: I'm not aware of other processes that could use them. I just limited to gen_addon_readme changes.

pedrobaeza:
This should be deployed on OCA server and then we should expect a massive README update.

PR #541, opened 2022-10-03
oca-create-branch: carry over answers from previous branches

sbidoul:
Carry over the repo name and repo description from previous branch.

### DIFF
--- a/tools/create_branch.py
+++ b/tools/create_branch.py
@@ -1,13 +1,32 @@
-"""Create a branch in all addons project.
-
-TODO
-- load copier answers from a previous branch
-"""
+"""Create a branch in all addons project."""
+from pathlib import Path
 import subprocess
+from typing import Dict
 
 import click
+import yaml
+import copier
 
 from .oca_projects import get_repositories, temporary_clone
+
+COPIER_ANSWERS_FILE = ".copier-answers.yml"
+COPIER_ANSWERS_TO_CARRY_OVER = ("repo_description", "repo_name")
+
+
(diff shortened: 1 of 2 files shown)


tools/create_branch.py
```diff
@@ -25,7 +48,11 @@
     "repos",
     multiple=True,
 )
-def main(new_branch, copier_template, copier_template_vcs_ref, repos):
+@click.option(
+    "--prev-branch",
+    help="Previous branch where to read some copier answers.",
```
simahawk, line 53: default True?

simahawk (commented):
LG, just a question

tools/create_branch.py
sbidoul: It is actually a branch name.

simahawk:
@sbidoul in the repo-maintainer tool will already preserve the repo name and description. Hence, I don't think we need this anymore.
See https://github.com/OCA/repo-maintainer/blob/master/oca_repo_maintainer/tools/manager.py#L191

Am I wrong?

Then we might implement a way to keep other keys but it can be done later.

sbidoul:
I think with repo_description I wanted to preserve the body of the repo README.md. But in practice very few repo have it in a copier answer.

So happy to drop oca-create-branch from here and use repo-maintainer-conf for that.

Thanks for that work!

sbidoul:
I keep this open for a while and if we are happy with the new solution, I'll remove create_branch.py from this repo.

sbidoul:
Closing since this is now handled by github.com/oca/repo-maintainer-conf

PR #545, opened 2022-11-11
[IMP] adapt to work on Python3.11

etobella:
Python3.11 no longer works with Universal newline in file read mode, as it is used by default from 3.3

### Milestone (Odoo version)
- 

### Module(s)
- 

### Fixes / new features
- Adapts the oca-gen-addon-readme to work with Python3.11. As you can see in https://docs.python.org/3.11/whatsnew/3.11.html
Universal file read mode flag is no longer supporte as it is the default after python 3.3

### DIFF
--- a/tools/gen_addon_readme.py
+++ b/tools/gen_addon_readme.py
@@ -2,7 +2,6 @@
 # Copyright (c) 2018 ACSONE SA/NV
 # Copyright (c) 2018 GRAP (http://www.grap.coop)
 
-import io
 import os
 import re
 import sys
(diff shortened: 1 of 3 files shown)


pedrobaeza (approved):

tools/gen_addon_readme.py
```diff
@@ -179,7 +182,7 @@ def gen_one_addon_readme(
             addon_dir, FRAGMENTS_DIR, fragment_name + '.rst',
         )
         if os.path.exists(fragment_filename):
-            with io.open(fragment_filename, 'rU', encoding='utf8') as f:
+            with io.open(fragment_filename, read_mode, encoding='utf8') as f:
```
sbidoul, line 185: Since the U mode was deprecated since python 3.3 you can simply drop the read_mode ('r' is the default).
BTW, while you are at it, you can replace `io.open` with `open`.

etobella: Sure, I wanted to keep the old compatibility, but you are right :smile:

etobella: Done! :smile:

sbidoul:
Thanks!

PR #557, opened 2023-02-13
[IMP] oca-gen-addon-readme: Allow to use custom templates

Shide:
### Milestone (Odoo version)
- all

### Module(s)
- oca-gen-addon-readme

### Fixes / new features
- Allow to use custom templates passing `--template-filename=/path/to/filename.template` to the **oca-gen-addon-readme** command

MT-303 @moduon @yajo

### DIFF
--- a/tools/gen_addon_readme.py
+++ b/tools/gen_addon_readme.py
@@ -170,8 +170,8 @@ def generate_fragment(org_name, repo_name, branch, addon_name, file):
     return fragment
 
 
-def gen_one_addon_readme(
-        org_name, repo_name, branch, addon_name, addon_dir, manifest):
+def gen_one_addon_readme(org_name, repo_name, branch, addon_name, addon_dir,
+                         manifest, template_filename):
     fragments = {}
     for fragment_name in FRAGMENTS:
         fragment_filename = os.path.join(
@@ -205,8 +205,6 @@ def gen_one_addon_readme(
         # maintainers section
     ]
     # generate
-    template_filename = \
-        os.path.join(os.path.dirname(__file__), 'gen_addon_readme.template')
     readme_filename = \
         os.path.join(addon_dir, 'README.rst')
     with open(template_filename, 'r', encoding='utf8') as tf:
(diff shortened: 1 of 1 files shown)


tools/gen_addon_readme.py
```diff
@@ -170,8 +170,8 @@ def generate_fragment(org_name, repo_name, branch, addon_name, file):
     return fragment
 
 
-def gen_one_addon_readme(
-        org_name, repo_name, branch, addon_name, addon_dir, manifest):
+def gen_one_addon_readme(org_name, repo_name, branch, addon_name, addon_dir,
+                         manifest, template_filename):
```
yajo, line 174: You can just use a default argument here:

```suggestion
def gen_one_addon_readme(org_name, repo_name, branch, addon_name, addon_dir,
                         manifest, template_filename=os.path.join(os.path.dirname(__file__), 'gen_addon_readme.template')):
```

tools/gen_addon_readme.py
```diff
@@ -205,7 +205,7 @@ def gen_one_addon_readme(
         # maintainers section
     ]
     # generate
-    template_filename = \
+    template_filename = template_filename or \
         os.path.join(os.path.dirname(__file__), 'gen_addon_readme.template')
```
yajo, line 209: .. and remove this

```suggestion
```

tools/gen_addon_readme.py
```diff
@@ -285,8 +285,10 @@ def gen_one_addon_index(readme_filename):
               help="git commit changes to README.rst, if any.")
 @click.option('--gen-html/--no-gen-html', default=True,
               help="Generate index html file.")
-def gen_addon_readme(
-        org_name, repo_name, branch, addon_dirs, addons_dir, commit, gen_html):
+@click.option('--template-filename', default=False,
```
yajo, line 288: I'm not sure if click can infer the default from the method signature, but if so:

```suggestion
@click.option('--template-filename',
```

tools/gen_addon_readme.py
Shide: Changed default value on  @click option

yajo (approved):

yajo:
@OCA/community-maintainers this one is an easy enhancement. Could somebody merge please?

tools/gen_addon_readme.py
dreispt: I'm pretty sure the hanging indentation is not pylint friendly.
Is this code covered by the same standards as Odoo module repos.

yajo: No, it has its own standards AFAICS: https://github.com/OCA/maintainer-tools/blob/e3b46970af62cbbf99f0b45a6675eb4cae744d6d/.pre-commit-config.yaml#L4-L9

When somebody adds black there, then all will be consistent. But that seems a separate issue, as the rest of the repo is already not black-friendly.

sbidoul:
LGTM.

PR #562, opened 2023-03-08
[IMP] gen_addon_readme: rename template and only advertise github urls for OCA

yajo:
Better syntax highlight.

Other orgs use other vcs trackers.

@moduon MT-956


yajo:
Any help with those CI ❌ ?

sbidoul:
> Any help with those CI

This are regular failures that seem related to the change in this PR.

sbidoul (changes requested):
There are organizations other than OCA for which these paragraphs are just fine. So I think this should be managed with a custom template.

Ok for the template renaming, though :+1:

yajo:
OK, I just kept the rename. Maintaining our own downstream template isn't a big burden anyways.

sbidoul:
@yajo I have included your renaming commit in #566

PR #566, opened 2023-04-01
oca-gen-addon-readme: add --if-fragments-changed option

### LINKED ISSUE #520: oca-gen-addon-readme: do not regenerate if content did not change
**Is your feature request related to a problem?**

Currently, the oca github bot generates all readme files nightly. This is good to provide uniformity and keep them fresh.
However, this mean that tiny changes in the readme generator output will regenerate thousands of README files, and publish thousands of packages to our wheelhouse and PyPI.

Consequently, we have [pinned](url) the version in the bot.

**Describe the solution you'd like**

Store in the generated README a rst comment with a hash of the content of the addon readme fragments.
Add an option to not regenerate the README if the hash did not change.

**Describe alternatives you've considered**

- generating readmes in ocabot merge command only, but this means the occasional manual merge will not generate the readme

**Additional context**

Add any other context about the feature request here.

sbidoul:
Don't generate `README.rst` if the fragments did not change, according to a digest stored as a comment in the target file.

This way we can avoid a mass-regeneration each time the template evolves.

closes #520 

Also, rename the template file, for better support (part of #562).


tools/gen_addon_readme.template
```diff
@@ -11,10 +11,13 @@
 {{ manifest.name }}
 {{ '=' * manifest.name|length }}
 
-.. !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+..
```
pedrobaeza, line 14: Why this extra line?

pedrobaeza:
Shouldn't be the option the contrary? By default, don't regenerate if no changes in the fragments, but if `--force` or similar is passed, then regenerate always.

tools/gen_addon_readme.template
sbidoul: Because this is the [correct syntax for multiline rst comments](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#comments).

sbidoul:
> Shouldn't be the option the contrary? By default, don't regenerate if no changes in the fragments, but if `--force` or similar is passed, then regenerate always.

My reasoning is to keep backwards compatibility and enable the new behaviour explicitly.
The pre-commit hook whould have `--if-fragments-changed` and when it is in pre-commit people would not need to run gen-addon-readme manually anyway so this should be fine?

pedrobaeza:
OK, I'm now thinking that when migrating, you need to regenerate it although the fragments are not changing, so it's sane to keep it that way as a residual option for the bot and those knowing what to do.

pedrobaeza (approved):

sbidoul:
We need to add `__manifest__.py` in the fragments hash, since manifest content influence the result.

yajo (approved):
LGTM once manifest is added to hashed files.

It should let us unlock https://github.com/OCA/oca-addons-repo-template/pull/190 which is an important issue for us. 

Thanks for the proposal!

PR #567, opened 2023-04-06
gen-addon-readme: use the canonical URL for runboat builds

sbidoul:
The other URL is an implementation detail of the current UI.


tools/gen_addon_readme.py
```diff
@@ -105,7 +105,7 @@
 def make_runboat_badge(repo, branch):
     return (
         "https://img.shields.io/badge/runboat-Try%20me-875A7B.png",
-        "https://runboat.odoo-community.org/webui/builds.html?"
+        "https://runboat.odoo-community.org/builds?"
```
yajo, line 108: It looks to me like this should be a macro in the template and not a function in the app. That doesn't make the fix incorrect though.

yajo (approved):

Issue #564, opened 2023-03-28
[RFC] Updating the README generator automation


sbidoul:
Currently, the `README.rst` in each addon is generated with the `oca-gen-addon-readme` by `oca-github-bot`.
The bot generates the README during "ocabot merge" and also on each direct push to the main branches.

The version of the script is pinned in the bot source code, because when we change it, it causes a massive regeneration of README files on all branches which in turns publishes all addons to PyPI and triggers a massive CI rebuild.

We need to unlock this situation so we can deploy the updated README generator.

Several approaches are being discussed:

1. Doing it in pre-commit and not with the bot (https://github.com/OCA/oca-addons-repo-template/pull/190)

This suffers the same drawback as the current situation, since each update to the repo template would trigger a massive re-generation.

2. Re-generating the README only when the content of the `readme/` directory changes (#520)

This might work but requires some work in the README generator.
When deployed there would be a massive re-generation, but that is reasonable if it is then stable.

3. Generating the README upon ocabot merge.

This requires some work in the bot, but is likely easier to implement than 2.
Ideally it should regenerate only for addons modified by the PR (the core logic is in place to detect that in the bot).
The README would be regenerated only when a PR is merged with the bot command.
It would not regenerate if there are direct pushes or manual merges.

Thoughts?

pedrobaeza:
Can't we freeze the versions in pre-commit config for being sure no change is done?

sbidoul:
> Can't we freeze the versions in pre-commit config for being sure no change is done?

Of course but it is not different from the current situation where the version is frozen in the bot.
The problem occurs when we want to upgrade the readme generator and we don't want to update 12 000 addons.

pedrobaeza:
But we control that with the copier template. We can apply such update only for the newest version, and reduce the impact. We are already doing this for black IIRC.

yajo:
It seems to me that the format of the `README.rst` is more or less stable, whereas the format of the `static/description/index.html` is more prone to changes depending on dependency versions. That is because `README.rst` just gets generated with a dumb jinja template, but `index.html` goes through a full rendering pipeline.

One simple guard would be to instruct `oca-gen-addon-readme` to not regenerate `index.html` if `README.rst` didn't change. Most times it won't change, so it should be just OK to leave it as a pre-commit hook and let the bot forget about readmes.

IIUC the bot is the same for all version branches. Removing the readme generation from the bot would free it from having to pin the maintainer-tools version.



[...]

PR #570, opened 2023-04-29
Unpin docutils

### LINKED ISSUE #423: readme generator generates different output depending on docutils version installed
Exemple:

```html
<meta name="generator" content="Docutils 0.15.1: http://docutils.sourceforge.net/" />
```

This is annoying because an accidental docutils upgrade (it is not pinned in maintainer-tools setup.py) can trigger a regenaration of all readme index.html, with the useless republishing of a ton of packages to pypi, and hours of travis builds.

First we should pin docutils to 0.15.1 in maintainer-tools setup.py with a comment explaining why.

This solution will break at some point because inevitably, in the future, that specific version will end up being not installable or too old for some other dependency...

sbidoul:
Since we now have `oca-gen-addon-readme --if-fragment-changed` which we will deploy to the OCA bot and pre-commit we can unpin docutils.

closes #423 
towards #564

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -24,10 +24,7 @@ setuptools.setup(
     install_requires=[
         "appdirs",
         "click",
-        # We need to pin docutils version, see
-        # https://github.com/OCA/maintainer-tools/issues/423
-        # Consider carefully before changing this.
-        "docutils==0.16.*",
+        "docutils",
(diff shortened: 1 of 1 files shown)


rvalyi (approved):

yajo (approved):
Makes sense.

PR #574, opened 2023-07-01
Add suport for markdown fragments in the README.rst generator

sbidoul:
This PR adds support for authoring readme fragments in markdown.

It also add an option to convert existing reStructuredText fragments to markdown.

The generated `README.rst` remains reStructuredText, as before.

The main goal is to allow for easier documentation contributions by non-technical contributors (markdown is a simpler format, and has better tooling support - in particular it allows easy image drag and drop in GitHub's markdown editor).


sbidoul:
@TumbaoJu

sbidoul:
If there are no objections, I plan to merge this soon-ish and do some deeper tests. 

This is totally backward compatible.

pedrobaeza (approved):

yajo:
Why is this lang translation needed? Wouldn't it be easier to render a README.md if fragments are on md, and a README.rst if they are on rst? 🤔 

After all, Odoo will only render the index.html anyways 🤷🏼‍♂️

yajo (approved):
My previous comment is non-blocking anyways 😉

pedrobaeza:
Please let the RST generation. What I would remove is the HTML generation that is ugly and bloated.

yajo:
If you want the README to be properly rendered on Github, the best is to have it on md. If you want it properly rendered on Odoo, the best is html. It seems to me like README.rst isn't really the best option for any of the main use cases we want to cover...

sbidoul:
Note that if you have `README.md` but no `index.html`, Odoo parses `README.md` as reStructuredText.

So for now my plan is to keep generating `README.rst`.

yajo:
> Note that if you have `README.md` but no `index.html`, Odoo parses `README.md` as reStructuredText.

Ah OK, then it makes sense. I thought that, if there's an `index.html`, Odoo forgot about any readmes.

sbidoul:
An example conversion of the fragments from rst to markdown, on a repo with extensive documentation, is visible in https://github.com/OCA/rest-framework/pull/370

yajo:
IMHO the template should include the fragments in markdown then.

PR #588, opened 2023-11-11
gen-addon-readme: add --keep-source-digest

sbidoul:
Add an option to not change the source digest in the generated file.

Since 
- we want to generate README.rst in pre-commit,
- the manifest is part of the source digest,
- the manifest can change even if PR does not modify the README, 

this option should help avoid merge conflicts.


pedrobaeza (approved):

legalsylvain (approved):
Thanks a lot !

Issue #595, opened 2023-12-08
oca-gen-addon-readme: stop emitting an XML declaration


sbidoul:
Since https://github.com/odoo/odoo/commit/8d06889ec3b4f7234ebe9e8f524b286e5dcffa29, Odoo emits a warning if index.html contains an xml declaration.

We should update the generator to not emit it, and also probably to generate html5 instead of xhtml. To be investigated.

sbidoul:
This was done in #596

PR #596, opened 2023-12-08
[IMP] oca-gen-addon-readme: stop emitting an XML declaration

nilshamerlinck:
https://github.com/OCA/maintainer-tools/issues/595

### DIFF
--- a/tools/gen_addon_readme.py
+++ b/tools/gen_addon_readme.py
@@ -111,6 +111,15 @@ RST2HTML_SETTINGS = {
     # Use the short form of syntax highlighting so that the generated
     # Pygments CSS can be used to style the output.
     "syntax_highlight": "short",
+    # Since odoo/odoo@8d06889, Odoo emits a warning
+    # if index.html contains an xml declaration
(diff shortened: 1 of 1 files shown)


pedrobaeza:
But this should be only for 17.0, not the rest of the versions.

bosd (approved):
LGTM, dunno if it is a blocker to do this for all versions

nilshamerlinck:
Hello guys :)
- thanks for looking into this PR!
- it's ok to apply this change for all versions: it will not change anything even for old odoo instances that would not have been patched with the fixes https://github.com/odoo/odoo/pull/133708 and https://github.com/odoo/odoo/pull/133900
- that's because we never actually strictly needed the XML declaration, because `docutils` also adds this meta tag in the `index.html`: `<meta http-equiv="Content-Type" content="text/html; charset=utf-8" />` that is respected by the `HTMLParser` used by `lxml.html.document_fromstring(doc)`
- I updated the commit to make this a bit more explicit

sbidoul:
Thanks!

PR #601, opened 2024-01-19
[FIX] do not use MAIN_BRANCHES anymore

remi-filament:
### Fixes
This PR tends to fix the issue that 17.0 branch are not taken into account by oca_projects.py get_repositories_and_branches function causing for instance that these are not created in Weblate (see https://odoo-community.org/groups/contributors-15/contributors-1545495).

The script used for populating weblate is calling that function which does not return v17.0 branch (https://github.com/OCA/oca-weblate-deployment/blob/75ad176738d15ac2d99aaf0311e31036e39e74d9/wocg-oca#L12C21-L12C50) 

According to https://github.com/OCA/maintainer-tools/pull/539 MAIN_BRANCHES was deprecated and should not be used anymore, so I remove it from default parameter.

With this change, 17.0 branches are properly returned.

### DIFF
--- a/tools/config.py
+++ b/tools/config.py
@@ -69,21 +69,5 @@ NOT_ADDONS = {
 }
 
 
-# deprecated, use is_main_branch() instead
-MAIN_BRANCHES = (
-    "6.1",
-    "7.0",
-    "8.0",
-    "9.0",
-    "10.0",
-    "11.0",
-    "12.0",
-    "13.0",
-    "14.0",
-    "15.0",
-    "16.0",
-)
-
-
 def is_main_branch(branch):
     return re.match(r"^(6\.1|\d+\.0)$", branch)
--- a/tools/oca_projects.py
+++ b/tools/oca_projects.py
@@ -18,7 +18,7 @@ import subprocess
 import tempfile
 
 import appdirs
-from .config import MAIN_BRANCHES, NOT_ADDONS
+from .config import NOT_ADDONS, is_main_branch
 from .github_login import login
 
 ALL = ["OCA_PROJECTS", "OCA_REPOSITORY_NAMES", "url"]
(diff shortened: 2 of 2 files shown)


remi-filament:
Hi @sbidoul do you mind having a look at this PR that should fix the missing 17.0 branches in Weblate ?

Best Regards !

sbidoul:
Thanks for taking care about this.

I worry this change will break things, as it will expose non-main branches such as the merge bot branches.

Would it make sense to pass `branches=()` in the weblate script, together with a branch_filter function ?
There is such a filter in the bot: https://github.com/OCA/oca-github-bot/blob/a364253511c15a6d7e0bf6e60cba973481fb9cbd/src/oca_github_bot/version_branch.py#L20-L26

remi-filament:
I have added the use of is_main_branch function in branch_filter by default in the function.
Is that not the purpose of this is_main_branch to ensure that only main branches are taken into account ?

I do not see that much of a risk, but I do not have your level of knowledge so I trust you on that !
Should I update the list of version to add 17.0 in config.py instead ?

sbidoul:
Ah yes, I see it now :facepalm: 

And the MAIN_BRANCHES global is already deprecated. So let's remove it completely if we stop maintaining it?

While you are at it let's drop 8, 9, 10 [here](https://github.com/OCA/oca-weblate-deployment/blob/75ad176738d15ac2d99aaf0311e31036e39e74d9/wocg-oca#L9) ?

remi-filament:
Alright, I just removed MAIN_BRANCHES variable from config.py as well !
Sure, I will update the weblate deployment repo as well.

sbidoul (approved):

sbidoul:
BTW (not for this PR), we can also drop OCA_PROJECTS, since it is completely out of date.
The source of truth for this (if one does not want to use the GitHub API) is in https://github.com/OCA/repo-maintainer-conf.

remi-filament:
> BTW (not for this PR), we can also drop OCA_PROJECTS, since it is completely out of date. The source of truth for this (if one does not want to use the GitHub API) is in https://github.com/OCA/repo-maintainer-conf.

I was looking into it, but OCA_PROJECTS is used in check_contrib script.
Before trying to update check_contrib to get it working with OCA_PROJECTS, maybe you can confirm that check_contrib is still needed (otherwise I can simply remove that script as well) ?

sbidoul:
> confirm that check_contrib is still needed

I don't know. Never used it myself. You can perhaps simply replace OCA_PROJECT by get_repositories() in that script.

PR #607, opened 2024-03-01
[FIX] oca-towncrier: Add support for Mardown format

lmignon:
If readme fragments uses the _md_ format, generates the history file into the same format

This should fix the issue preventing the merge of some PR as https://github.com/OCA/search-engine/pull/176#issuecomment-1857431949


ivs-cetmix:
@lmignon thank you for this PR! We have exactly the same issue in our local repos after wi switched to .md format and were looking for solution.

lmignon:
ping @sbidoul

sbidoul:
Thank you. This is deployed on the OCA github bot.

PR #620, opened 2024-07-04
oca-gen-addon-readme: delete pandoc installer

### LINKED ISSUE #619: pandoc-3.2-1-amd64.deb added by precommit hook
Is it intended that the precommit hooks add [pandoc-3.2-1-amd64.deb](https://github.com/OCA/operating-unit/pull/667/files#diff-c64b58dd76f52ba2c9e8b7d0818f8c4d2e9f732788ec8b650df70762c852ddd1) to the repositories? 

I've seen this multiple time in PR reviews, and I'm wondering if this is intentional, or a blooper.

nilshamerlinck:
- when pandoc is not available in PATH, [`ensure_pandoc_installed`](https://github.com/JessicaTegner/pypandoc/blob/1341920fa1c1ce661a0ac22138ce819eb73d9f77/pypandoc/__init__.py#L821) downloads its installer in current directory and unpacks it
- one might question that intrusive behavior; in our case we can consider it as beginner friendly
- we just need to delete the installer to avoid polluting repos  
 
fixes #619

### DIFF
--- a/tools/gen_addon_readme.py
+++ b/tools/gen_addon_readme.py
@@ -131,7 +131,7 @@ PANDOC_MARKDOWN_FORMAT = "gfm-raw_html-gfm_auto_identifiers"
 
 @functools.lru_cache(maxsize=None)
 def ensure_pandoc_installed() -> None:
-    pypandoc.ensure_pandoc_installed()
+    pypandoc.ensure_pandoc_installed(delete_installer=True)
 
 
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):
Thanks for the patch. It has been a total pain.

sbidoul:
Thanks!

PR #624, opened 2024-07-11
Update USAGE.md

TumbaoJu:
Proposition from functional working group

### DIFF
--- a/template/module/readme/USAGE.md
+++ b/template/module/readme/USAGE.md
@@ -12,7 +12,7 @@ To use this module, you need to:
 
   *insert screenshot!*
 
-- In “Contact” form, add a value to field `xyz` > save
+- In “Contact” form, add a value to field *xyz* > save
 
   *insert screenshot!*
 
--- a/template/module/readme/USAGE.md
+++ b/template/module/readme/USAGE.md
@@ -1,11 +1,21 @@
-[ This file must be present and contains the usage instructions
-  for end-users. As all other rst files included in the README,
-  it MUST NOT contain reStructuredText sections
-  only body text (paragraphs, lists, tables, etc). Should you need
-  a more elaborate structure to explain the addon, please create a
-  Sphinx documentation (which may include this file as a "quick start"
-  section). ]
+[ This file is required and contains the instructions on **“how”** to use the module for end-users. 
+
+If the module does not have a visible impact on the user interface, just add the following sentence: 
+
+> This module does not impact the user interface.
+
+If that’s not the case, please make sure that every usage step is covered and remember that images speak more than words!]
 
 To use this module, you need to:
 
-1. Go to ...
+- Go to *App* > Menu > Menu item
+
+  *insert screenshot!*
+
+- In “Contact” form, add a value to field `xyz` > save
+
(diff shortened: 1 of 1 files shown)


template/module/readme/USAGE.md
```diff
@@ -1,11 +1,21 @@
...
+If that’s not the case, please make sure that every usage step is covered and remember that images speak more than words!]
 
 To use this module, you need to:
 
-1. Go to ...
+- Go to *App* > Menu > Menu item
+
+  *insert screenshot!*
+
+- In “Contact” form, add a value to field `xyz` > save
```
yajo, line 15: suggestion: imho we should use code markers only when referring to code:

```suggestion
- In “Contact” form, add a value to field *xyz* > save
```

template/module/readme/USAGE.md
```diff
@@ -1,11 +1,21 @@
...
+
+  *insert screenshot!*
+
+- The value of `xyz` is now displayed in the list view.
```
yajo, line 19:
```suggestion
- The value of *xyz* is now displayed in the list view.
```

yajo (approved):

francesco-ooops (approved):

rafaelbn:
Hello @TumbaoJu , Could you please review the comments from @yajo? Thank you!

template/module/readme/USAGE.md
TumbaoJu: @yajo Usually, the field names are formatted in `code` so they are easier to see into the documentation.

TumbaoJu:
@OCA/oca-consultants : Here is one of the PR that propose changes into the description of the fragments from the Read Me template.

template/module/readme/USAGE.md
francesco-ooops: @TumbaoJu tbh I haven't seen this often in current documentation, is there any guidelines suggesting so?

TumbaoJu: @francesco-ooops : Maybe it is just in Numigi? I always though it was the basic but maybe not.
In my opinion, it is easier to read a documentation that uses the `code` instead of the *italic* format for field names.

Maybe we could make it a new guideline?

What do you think?

TumbaoJu: Here is an example : https://github.com/OCA/web/tree/14.0/web_custom_modifier

francesco-ooops: Seems like field labels are in italics, eg: 

> As system administrator, I go to _Settings / Technical / User Interface / Custom Modifiers_.

While technical names of fields are indeed using `code`

So I would keep italics for label names :)

yajo: Yes, that's it. IMHO this sentence would be properly formatted: "The field `order_line_id` is labeled *Order Lines*".

Since the documentation is functional, I think most of the times you'll write field labels and not field technical names. That's why I asked for this change.

`web_custom_modifier` is an exception because it requires you to input the technical name of the field. In those cases, code formatting is appropriate.



> Maybe we could make it a new guideline?

Please do! 😊

yajo (approved):

PR #629, opened 2024-09-26
Update ocb-sync for Odoo 18

sbidoul:


pedrobaeza:
I think we can disable the 14.0 sync.

sbidoul:
Probably, yes, I agree.

pedrobaeza:
Done on #630

PR #631, opened 2024-10-01
oca-towncrier: Use correct title level

lmignon:
Use of the correct title level when generating the history into md format

This is required to fix https://github.com/OCA/rest-framework/pull/453#issuecomment-2385509745

ping @sbidoul


sbidoul:
Deployed in the OCA github bot.

PR #638, opened 2025-01-31
[DOC] mention missing context.rst file

legalsylvain:
ref https://github.com/OCA/pos/pull/1278#issuecomment-2626629987

CC : @hbrunn

### DIFF
--- a/template/module/README.rst
+++ b/template/module/README.rst
@@ -5,7 +5,7 @@
 Please provide content in the ``readme`` directory:
 
 * **DESCRIPTION.rst** (required)
-* CONTEXT.rst (optional)
+* CONTEXT.rst (optional, strongly recommended)
 * INSTALL.rst (optional)
 * CONFIGURE.rst (optional)
 * **USAGE.rst** (optional, highly recommended)
--- a/template/module/README.rst
+++ b/template/module/README.rst
@@ -5,6 +5,7 @@
 Please provide content in the ``readme`` directory:
 
 * **DESCRIPTION.rst** (required)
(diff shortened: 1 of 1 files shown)


hbrunn (approved):

rousseldenis (approved):

template/module/README.rst
```diff
@@ -5,6 +5,7 @@
 Please provide content in the ``readme`` directory:
 
 * **DESCRIPTION.rst** (required)
+* CONTEXT.rst (optional)
```
sbidoul, line 8:
```suggestion
* CONTEXT.rst (optional, strongly recommended)
```

ivs-cetmix: > I dont see this file very often. Why do you see it as important ?

Actually this file is not important. It's critical for understanding the context of the module and it's really precious when it comes to the functional part, especially module applicability.
I would even opt it to be mandatory, however it may be an overkill for some purely technical modules.
Yet for most of the module it's a must.
Here is just one [example](https://github.com/OCA/stock-logistics-warehouse/blob/16.0/stock_picking_product_interchangeable/readme/CONTEXT.md).

sbidoul: CONTEXT is relatively new and not well known yet, but it is indeed important to encourage people to document the use case of the modules. Currently with DESCRIPTION and USAGE, we see what and how, but it's often difficult to understand why/when a module is useful.

sbidoul:
Thanks!

BTW, for 17+ it's now markdown. I've not looked if the template needs to be adapted for that though.

PR #642, opened 2025-02-06
[IMP] readme model updated with oca img banner

dora-jurcevic:
Readme model modified in order to display a OCA  img banner on the top with a link inviting users to become members or sponsors

### DIFF
--- a/tools/gen_addon_readme.rst.jinja
+++ b/tools/gen_addon_readme.rst.jinja
@@ -7,6 +7,11 @@
 {% endif %}
 {%- endmacro -%}
 
+.. image:: https://odoo-community.org/web/image/4603934-dcc7fad0/readme_banner_OCA.png
+   :target: https://odoo-community.org/get-involved#scrollTop=0?ref=u3kel3&utm_source=oca-readme-banner-module&utm_medium=referral&utm_campaign=become-member-2025
+   :alt: Odoo Community Association
+
+
 {{ '=' * manifest.name|length }}
 {{ manifest.name }}
 {{ '=' * manifest.name|length }}
--- a/tools/gen_addon_readme.rst.jinja
+++ b/tools/gen_addon_readme.rst.jinja
@@ -7,10 +7,12 @@
 {% endif %}
 {%- endmacro -%}
 
-.. image:: https://odoo-community.org/web/image/4603934-dcc7fad0/readme_banner_OCA.png
-   :target: https://odoo-community.org/get-involved#scrollTop=0?ref=u3kel3&utm_source=oca-readme-banner-module&utm_medium=referral&utm_campaign=become-member-2025
-   :alt: Odoo Community Association
+=======
+.. image:: https://odoo-community.org/web/image/4610395-7254fada/OCA_banner_readme_1256x150px.png
+   :target: https://odoo-community.org/get-involved?ref=k8vx13&utm_source=referral-readme&utm_medium=referral
+   :alt: Odoo Community Association 
 
+=======
 
 {{ '=' * manifest.name|length }}
 {{ manifest.name }}


tools/gen_addon_readme.rst.jinja
```diff
@@ -7,6 +7,11 @@
 {% endif %}
 {%- endmacro -%}
 
+.. image:: https://odoo-community.org/web/image/4603934-dcc7fad0/readme_banner_OCA.png
+   :target: https://odoo-community.org/get-involved#scrollTop=0?ref=u3kel3&utm_source=oca-readme-banner-module&utm_medium=referral&utm_campaign=become-member-2025
```
rousseldenis, line 11: This implies the campaign should remain open

dora-jurcevic: If you mean the OCA sponsors recruitment campaign, yes, it stays open

sbidoul: I think we need simpler urls and stable  that are not tied to odoo implementation details. Otherwise they will rapidly become broken as we upgrade Odoo etc...

vdewulf:
The URL was created with a separate too where I put the original link and
added details like the source and the campaign name (but I can remove 2025
for example) to analyse how many people visited the website after clicking
on this banner.

Can you explain how to change a URL in odoo so that it stays the same after
a migration?  And how I can use this separate tool so that it works as well?

Le ven. 7 févr. 2025, 08:38, Stéphane Bidoul ***@***.***> a
écrit :

tools/gen_addon_readme.rst.jinja
rousseldenis: > I think we need simpler urls and stable that are not tied to odoo implementation details. Otherwise they will rapidly become broken as we upgrade Odoo etc...

That was the feeling I had but @sbidoul said it clearly :-)

sbidoul:
> Can you explain how to change a URL in odoo so that it stays the same after
a migration?

We can probably do somethings with Website Settings > Configuration > Redirects.

sbidoul:
Another comment I have is about the banner. It looks very big and too "advertising".

If we see such a big banner on each addon page, either on pypi.org or our own app store, it will rapidly becoming tiring, I think.

See for instance how it looks now on https://pypi.org/project/odoo-addon-mis-builder/.

Can we do something less intrusive?

dora-jurcevic:
> Another comment I have is about the banner. It looks very big and too "advertising".
> 
> If we see such a big banner on each addon page, either on pypi.org or our own app store, it will rapidly becoming tiring, I think.
> 
> See for instance how it looks now on https://pypi.org/project/odoo-addon-mis-builder/.
> 
> Can we do something less intrusive?

So the starting point of the banner idea is the idea that many users use the OCA modules without even knowing that they can "give back" to the community by sponsoring it or becoming members. 
In the same time, the "refferal" url will give us the possibility to measure exactly how many people click on the banner and how many "convert" becoming members or sponsors. The idea is also to change this banner regularly, or even to make it evolve. We took as example other open source projects that live thanks to sponsors. 

I understand the fear of annoying publicity but  if it becomes annoying and in the same time we measure in 6-10 month that is inefficient, it will be a good reason to replace it or remove it completely.

sbidoul:
Yes I understand the goal, but I'm pretty sure that using PyPI as an advertising medium would be frowned upon.

We can achieve the same result with something a more discreet.

dora-jurcevic:
> Yes I understand the goal, but I'm pretty sure that using PyPI blatantly as advertising medium would be frowned upon.
> 
> We can achieve the same result with something a more discreet.

ok, I understand. 
Is there a way to conditionally hide the banner on  PyPI  ? 
If the answer is no, can we include a "promotional message" in text format, with a link to odoo-community website ? Would this be acceptable ?

dora-jurcevic:
Hello @sbidoul @vdewulf @rousseldenis ! 
We worked on a smaller version of the banner and a more simple referal link.
Please check if this is more suitable for PyPI. 
Thx :)

JayVora-SerpentCS:
Its red, but good to go for merge!

vdewulf:
@etobella Just lookinf for more eyes for this PR to get merged... could you share your feedback with us?
Thanks!

rousseldenis (approved):

etobella (commented):
The change itself looks good, but tests are failing

vdewulf:
> The change itself looks good, but tests are failing

Thanks Enric! Indeed it is red, but @dora-jurcevic doesn't know how to make it green and @JayVora-SerpentCS said it was good to merge even if it is red.

What is the best option to move forward?

tools/gen_addon_readme.rst.jinja
```diff
@@ -7,6 +7,13 @@
 {% endif %}
 {%- endmacro -%}
 
+=======
+.. image:: https://odoo-community.org/web/image/4610395-7254fada/OCA_banner_readme_1256x150px.png
```
sbidoul, line 11: Could we use a nice stable URL here and set up a redirect on the website? In case our website changes we would not want to break existing READMEs.

tools/gen_addon_readme.rst.jinja
```diff
@@ -7,6 +7,13 @@
...
+   :target: https://odoo-community.org/get-involved?ref=k8vx13&utm_source=referral-readme&utm_medium=referral
```
sbidoul, line 12: What is "ref=k8vx13" ? Also here should we not use a stable URL in case the details of our website change in the future?

sbidoul (commented):
A couple of comments about stable URLs.

Can you also update the README.expected* files in https://github.com/OCA/maintainer-tools/tree/master/tests/data to fix the tests?

And rebase to get the latest CI updates and fixes.

tools/gen_addon_readme.rst.jinja
dora-jurcevic: Hello @sbidoul ! 
This URL was created when we uploaded the image png on the current OCA Website.
In order to have a stable URL for this image I suppose that we would need a module on the current Odoo ERP that creates a route with for example  https://odoo-community.org/oca-readme-banner.png ?
Or is there a more simple way of adding a media on a specific URL in Odoo V14?
Thx !

tools/gen_addon_readme.rst.jinja
dora-jurcevic: @sbidoul , here the base URL will be stable https://odoo-community.org/get-involved and we should update each year the referal part of the URL (?ref=k8vx13&utm_source=referral-readme&utm_medium=referral) in the readme repo. If this is not possible (updating the maintainer-tools each year), we should see with @vdewulf to drop the referal part of the url and keep only the base part of the url. This referal part is used for tracking info about the visits on the /get-involved page, to be sure that the visit comes from the readme banner. 
@sbidoul, do you think we should avoid updating URL each year ?

sbidoul:
I'm finishing this in #656

PR #649, opened 2025-04-16
gen-addon-readme: consistency between words "license" and "licence"

luisg123v:
Before this commit, "license" was used in:
- Image's alternative text
- License URL.

While "licence" was only used in the badge's text.

To be consistent, "license" is now used everywhere. We're preferring that form because:
- It's the way licenses are referred to in their actual names and content (e.g. "GNU Affero General Public License").
- License file is called `LICENSE`.
- The Manifest key of Odoo modules is called that way.

### Milestone (Odoo version)
- 

### Module(s)
- 

### Fixes / new features
-

### DIFF
--- a/tests/test_repo/__unported__/module3/README.rst
+++ b/tests/test_repo/__unported__/module3/README.rst
@@ -1,4 +1,4 @@
-.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+.. image:: https://img.shields.io/badge/license-AGPL--3-blue.svg
    :target: https://www.gnu.org/licenses/agpl
    :alt: License: AGPL-3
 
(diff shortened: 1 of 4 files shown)


rvalyi (approved):

StefanRijnhart (approved):

rousseldenis:
> Thanks for checking this. We should be aware that this will mean that every README in OCA modules since this landed (~15.0) will modify their README with this change the next time it's touched in the files of the checksum. Is this what we want?

I think this is not a drawback and can be part of modules enhancements as others.

rousseldenis (approved):

etobella (approved):
@pedrobaeza The change will not happen on real life because the pre-commit is associated to a specific commit, this will not be executed unless we do a major change on the copier template.

PR #650, opened 2025-04-21
Oca top banner test

dora-jurcevic:
Test for checking if creating a new PR can solve the issue of CI / Test with python 3.11 failing on my previous PR https://github.com/OCA/maintainer-tools/pull/642

### DIFF
--- a/tools/gen_addon_readme.rst.jinja
+++ b/tools/gen_addon_readme.rst.jinja
@@ -6,7 +6,12 @@
 {{ fragments[name] }}
 {% endif %}
 {%- endmacro -%}
+=======
+.. image:: https://odoo-community.org/web/image/4610395-7254fada/OCA_banner_readme_1256x150px.png
+   :target: https://odoo-community.org/get-involved?ref=k8vx13&utm_source=referral-readme&utm_medium=referral
+   :alt: Odoo Community Association 
 
+=======
 {{ '=' * manifest.name|length }}
 {{ manifest.name }}
 {{ '=' * manifest.name|length }}


ivs-cetmix:
Hello @sbidoul, I'm struggling to understand why exactly does this test fail. Would appreciate your assistance.
It's also weird that another PR with just a test file added failed as well: https://github.com/OCA/maintainer-tools/pull/650

tools/gen_addon_readme.rst.jinja
```diff
@@ -6,7 +6,12 @@
 {{ fragments[name] }}
 {% endif %}
 {%- endmacro -%}
+=======
+.. image:: https://odoo-community.org/web/image/4610395-7254fada/OCA_banner_readme_1256x150px.png
```
sbidoul, line 10: Could we use a nice stable URL here and set up a redirect on the website? In case our website changes we would not want to break existing READMEs.

tools/gen_addon_readme.rst.jinja
```diff
@@ -6,7 +6,12 @@
...
+   :target: https://odoo-community.org/get-involved?ref=k8vx13&utm_source=referral-readme&utm_medium=referral
```
sbidoul, line 11: What is "ref=k8vx13" ? Also here should we not use a stable URL in case the details of our website change in the future?

sbidoul (commented):
A couple of comments about stable URLs.

Can you also update the `README.expected*` files in https://github.com/OCA/maintainer-tools/tree/master/tests/data to fix the tests?

And rebase to get the latest CI updates and fixes.

sbidoul:
Ah, sorry, this is not the original PR. Let's close this one, I'll re-do my comments on the other one.

PR #656, opened 2025-06-03
gen-addon-readme: add OCA banner image

sbidoul:


sbidoul:
Fast tracking as this was previously discussed in #642

PR #661, opened 2025-09-15
Add 19.0 on OCB Sync

etobella:
Add 19 on the OCB Sync executable


tests/test_gen_addon_readme.py
```diff
@@ -162,7 +162,10 @@ def test_rst_error(tmp_path):
                 cmd + opts, stderr=subprocess.STDOUT, universal_newlines=True
             )
         except subprocess.CalledProcessError as e:
-            assert "Title level inconsistent" in e.output
+            assert (
```
etobella, line 165: I am not happy with this change, but different errors were raised for each python version 😢

tools/ocb-sync.sh
```diff
@@ -8,7 +8,7 @@
 
 ODOO="git@github.com:/odoo/odoo.git"
 OCB="git@github.com:/OCA/OCB.git"
-BRANCHES="18.0 17.0 16.0 15.0"
+BRANCHES="19.0 18.0 17.0 16.0 15.0"
```
sbidoul, line 11: Let's remove 15? it's very unlikely to evolve by now.

etobella: Well, Odoo will start maintaining 5 versions, isn't it?

sbidoul:
Deployed.

Note to self: there is a `ocb-sync.sh.conf` config file that overrides the script default.

PR #676, opened 2026-05-20
gen-addon-readme: use addon manifest name as HTML <title>

dnplkndll:
The OCA banner image at the top of every `README.rst` suppresses docutils' `doctitle_xform`, so the generated `static/description/index.html` falls back to using the input filename — every OCA module ships a static description page titled "README.rst". Read the addon's manifest `name` and pass it via `settings_overrides.title`.

Verified locally with `base_m2m_custom_field`: `<title>` becomes "Base Many2many Custom Field" instead of "README.rst".

### DIFF
--- a/tools/gen_addon_readme.py
+++ b/tools/gen_addon_readme.py
@@ -403,7 +403,7 @@ def check_rst(readme_filename):
         )
 
 
-def gen_one_addon_index(readme_filename):
+def gen_one_addon_index(readme_filename, addon_title=None):
     addon_dir = os.path.dirname(readme_filename)
     index_dir = os.path.join(addon_dir, "static", "description")
     index_filename = os.path.join(index_dir, "index.html")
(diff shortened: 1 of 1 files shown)


dnplkndll:
@sbidoul @petrus-v — quick docutils setting override; verified locally

sbidoul:
Nice!

StefanRijnhart (approved):
Thanks! Tested locally.

sbidoul:
Now the repo template needs updating.

I'll update the GitHub bot.

Next week, I'll also try to progress with automatic updates of the repo template with renovate.

pedrobaeza:
It's important to sync the copier template with what you deploy in the bot, because if not, we have a PR changing anything in the manifest, that regenerates the README.rst / index.html via pre-commit, adding a lot of noise in the PR, and then when merged, triggering another commit "post-merge updates" restoring again the files to the original state.

Although each repo doesn't have the latest copier template, at least to have the copier template definition synced, and then we can put manual updates per repo when we detect these situations.

sbidoul:
Yes I know it's annoying. I'm still not convinced having the README generation in pre-commit is a good idea.

pedrobaeza:
Please then sync both. I think it's very valuable to have the README generated for reviewers to have that file to read, and remember also the problem that the README is not correct when merging the fragments, that is only then detected on merge, not when running pre-commit.

